PR #18, opened 2014-07-02
[winrt support] Enabling Boost.Thread to be used in the Windows Runtime.

stgates:
This involves basically 3 changes:
1. Using __declspec(thread) instead of the Tls APIs.
2. Using Windows::System::Threading since Win32 Threading APIs aren't allowed.
3. Updating or replacing some banned APIs like WaitForSingleObject with WaitForSingleObjectEx.


jgleitsmann:
Would these versions be compatible with or useable for developing an Xbox app that needs Boost for threads, signals, ...?

stgates:
@jgleitsmann I'm not very familiar for developing for the Xbox, but potentially it could work.

PR #30, opened 2014-11-14
prevent ~unique_lock from throwing exceptions

tempoz:
~unique_lock() calls unlock(), which can throw an exception. Destructors may not throw exceptions on pain of immediately terminating the program, so we have to catch the exception.

### DIFF
--- a/include/boost/thread/lock_types.hpp
+++ b/include/boost/thread/lock_types.hpp
@@ -328,7 +328,11 @@ namespace boost
     {
       if (owns_lock())
       {
-        m->unlock();
+        try {
+          m->unlock();
+        } catch (boost::lock_error) {
+          //destructors may not throw
+        }
       }
     }
     void lock()


tempoz:
Here is the function body of `unlock()`:

```cpp
void unlock()
{
    int res = posix::pthread_mutex_unlock(&m);
    if (res)
    {
        boost::throw_exception(lock_error(res,"boost: mutex unlock failed in pthread_mutex_unlock"));
    }
}
```

So `unlock()` throws when `posix::pthread_mutex_unlock` returns any non-zero value. It is not my goal to hide this by catching the exception, just to prevent the entire program from violently terminating. If someone wanted to add logging in the catch block, I would not be opposed.

viboes:
There is no valid reason a call to `mutex::unlock()` inside a `unique_lock` throws.

```


    The pthread_mutex_lock() and pthread_mutex_trylock() functions shall fail if:

    [EINVAL]
        The mutex was created with the protocol attribute having the value PTHREAD_PRIO_PROTECT and the calling thread's priority is higher than the mutex's current priority ceiling.

    The pthread_mutex_trylock() function shall fail if:

    [EBUSY]
        The mutex could not be acquired because it was already locked.

    The pthread_mutex_lock(), pthread_mutex_trylock(), and pthread_mutex_unlock() functions may fail if:

    [EINVAL]
        The value specified by mutex does not refer to an initialized mutex object.
    [EAGAIN]
        [XSI] [Option Start] The mutex could not be acquired because the maximum number of recursive locks for mutex has been exceeded. [Option End]

    The pthread_mutex_lock() function may fail if:

    [EDEADLK]
        A deadlock condition was detected or the current thread already owns the mutex.

    The pthread_mutex_unlock() function may fail if:

    [EPERM]
        The current thread does not own the mutex.

    These functions shall not return an error code of [EINTR].
```

viboes:
What I could/should do is to don't throw on `Mutex::unlock()` . From the documentation

```
m.unlock();

Requires:

    The current thread owns m. 
Synchronization:

    This operation synchronizes with subsequent lock operations that obtain ownership on the same object. 
Effects:

    Releases a lock on m by the current thread. 
Return type:

    void. 
Throws:

    Nothing. 
```

The standard requires the same. I will create a Trac ticket.

tempoz:
Thank you; this solves my problem as well.

PR #40, opened 2015-01-27
Disable thread safety analysis / avoid confusing clang on OS X

jim-wong:
Original description, from thughes:

This particular path confuses clang on OSX (seems to work fine with clang 3.5 on Linux).


viboes:
Could you add the kind of errors this patch intends to fix?

viboes:
Closed as there is no response

PR #41, opened 2015-01-28
Moved function calls out of the BOOST_VERIFY macro

tempoz:
Due to the structure of some of the "assert" macros used by BOOST_VERIFY, it is unsafe to pass it function calls that are not idempotent (and thus good practice to not pass it function calls in general). In this particular instance, the pthread_mutex_lock call would be duplicated on a non-zero return from pthread_mutex_lock, resulting in a possible double lock. This issue was uncovered with Coverity.

### DIFF
--- a/include/boost/thread/pthread/pthread_mutex_scoped_lock.hpp
+++ b/include/boost/thread/pthread/pthread_mutex_scoped_lock.hpp
@@ -23,11 +23,13 @@ namespace boost
             explicit pthread_mutex_scoped_lock(pthread_mutex_t* m_):
                 m(m_),locked(true)
             {
-                BOOST_VERIFY(!pthread_mutex_lock(m));
+                int error_code = pthread_mutex_lock(m);
+                BOOST_VERIFY(!error_code);
             }
             void unlock()
             {
-                BOOST_VERIFY(!pthread_mutex_unlock(m));
+                int error_code = pthread_mutex_unlock(m);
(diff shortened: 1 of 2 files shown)


viboes:
Do you have an example of assert an implementation that is in this case?

pdimov:
If this is true, it's a bug in BOOST_VERIFY and needs to be fixed, so I'll appreciate more information about it.

pdimov:
Note that verify_test.cpp already tests `BOOST_VERIFY( ++x )` and `BOOST_VERIFY( f(x) )`, where `f` is not idempotent (it increments `x`).

tempoz:
yes, /usr/include/assert.h on my machine includes
'''
# define assert(expr)                                                   \
((expr)                                                               \
? __ASSERT_VOID_CAST (0)                                             \
: __assert_fail (__STRING(expr), __FILE__, __LINE__, __ASSERT_FUNCTION))
'''

pdimov:
This only evaluates `expr` once. I'm willing to bet that `__STRING(expr)` does not evaluate `expr`, it just turns it into a string. I don't know why `#expr` is not used directly.

tempoz:
My apologies, you appear to be correct. I had actually not encountered stringification before, and so did not recognize it as an instance in which expr would not be evaluated. Marking as false positive in Coverity, and not bothering you about it anymore :P

viboes:
Thanks.

PR #43, opened 2015-02-02
Reverted parts of c16ec42 because 45510fa was a better fix.

vtnerd:
Let me know if there are more tests I should run for verification.


viboes:
Thanks

PR #44, opened 2015-02-04
Win32 coalesced timers support

ned14:
Related to issue #9856.

Doesn't fail any timer related unit tests. BTW Vicente your when_any testing is failing heavily with bad iterator dereferencing. I'll email you the failure log.


viboes:
I don't master the windows interface. 
How this patch improves the current situation? Does it ensure that there are less than 10% of spurious notifications?
I let you do the merge as I can not test it.

ned14:
To be honest, this patch makes very little difference to issue #9856, about the same number of timed waits return early as before for short duration waits. I marked that issue as wontfix, because if Windows is returning early we can't do much but pass that on.

What it gains though is much better power consumption on mobile devices as the system can choose to extend the timed wait by up to 5%. And for longer waits e.g. 1 second, I've noticed Windows is almost exactly accurate to your request now whereas before it was going on average +30% over the requested amount.

PR #49, opened 2015-02-18
Added an alternative implementation to the serial_executor, which doesn't need an additional thread

frascm:
everything as intended?


viboes:
NOT ACCEPTED

PR #52, opened 2015-04-15
 Consistent Android build by including limits.h

PhilLab:
Including <limits.h> instead of manually defining PAGE_SIZE.

The old way of including <asm/page.h> was made unavailable since https://android-review.googlesource.com/#/c/83299/ and fixed by defining PAGE_SIZE manually. The question arises whether it would be more consistent to include limits.h which I hereby put up for discussion.

### DIFF
--- a/include/boost/thread/pthread/thread_data.hpp
+++ b/include/boost/thread/pthread/thread_data.hpp
@@ -25,9 +25,7 @@
 #include <utility>
 
 #if defined(__ANDROID__)
-# ifndef PAGE_SIZE
-#  define PAGE_SIZE 4096
-# endif
+#include <limits.h>
 #endif
 
 #include <pthread.h>


viboes:
thanks. I can accept only PR for the develop branch.

ned14:
Besides I don't think limits.h worked in older Android APIs, which is precisely why we hard code it.

I have no objection to using limits.h on Android 5.0 or later.

PhilLab:
When looking in my Android ndk r10d, limits.h is included in all platforms/android-<version>/<architecture>/usr/include so maybe this depends on the NDK version, not the Android version.

Concerning the development branch: ok, I will do so as soon as the "limits.h on older Android APIs" thing is resolved.

ned14:
I believe it is an NDK version thing. Which can be detected using an API version check.

I'll put it another way - no Android device to date doesn't use 4Kb pages, so therefore any which does must use Android 5.0 or later.

PhilLab:
Ok, then we best leave everything as it is before it gets too complicated. After all, it is only a minor thing.

PR #53, opened 2015-04-22
Remove unused parameter

tempoz:
lk is unused in this function, naming it in the definition causes build warnings.


viboes:
Thanks for reporting. Please send PR for develop.

viboes:
Already fixed in develop.

PR #58, opened 2015-06-22
thread::try_join_until: Avoid busy wait if system clock changes

mikecrowe:
If system clock changes by an amount larger than the outstanding duration
according to the supplied clock then the code loops around again but
continues to pass a time point based on the original value of the system
clock to the system_clock variant of try_join.

If we're going to recalculate the outstanding duration in the loop then it
is necessary to get the current time according to the system clock to use
with this duration.

Signed-off-by: Mike Crowe <mac@mcrowe.com>

### DIFF
--- a/include/boost/thread/detail/thread.hpp
+++ b/include/boost/thread/detail/thread.hpp
@@ -485,9 +485,9 @@ namespace boost
         bool try_join_until(const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          system_clock::time_point     s_now = system_clock::now();
           bool joined= false;
           do {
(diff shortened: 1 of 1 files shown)


viboes:
Thanks for catching this issue.
I could merge it if you create a PR on develop. Otherwise I can just apply the patch manually as it is really simple.

mikecrowe:
https://svn.boost.org/trac/boost/ticket/11688

HTH.

viboes:
https://github.com/boostorg/thread/commit/971c028c9a72ab4de7776bcff9b76c7a7b4e5e81

PR #59, opened 2015-06-28
Shared timed mutex implementation

eldiener:
Updated to have shared_timed_mutex class in place of shared_mutex class as per the latest C14 standard. The shared_mutex class is a typedef to shared_timed_mutex, but should be deprecated in the thread docs so that eventually shared_mutex can be a separate non-timed shared_mutex when the C17 standard occurs.

I did not change the documentation so that the deprecation of shared_mutex can be documented and end-users encouraged to use shared_timed_mutex instead. I leave this for the thread developer(s) to do.

I also did not check for other Boost libraries using shared_mutex. But since shared_mutex is now always a typedef for shared_timed_mutex, all uses of shared_mutex should continue to work as before.

### DIFF
--- a/include/boost/thread/pthread/shared_mutex.hpp
+++ b/include/boost/thread/pthread/shared_mutex.hpp
@@ -26,7 +26,7 @@
 
 namespace boost
 {
-    class shared_mutex
+    class shared_timed_mutex
     {
     private:
         class state_data
@@ -170,13 +170,13 @@ namespace boost
 
     public:
 
-        BOOST_THREAD_NO_COPYABLE(shared_mutex)
+        BOOST_THREAD_NO_COPYABLE(shared_timed_mutex)
 
-        shared_mutex()
+        shared_timed_mutex()
         {
         }
 
-        ~shared_mutex()
+        ~shared_timed_mutex()
         {
         }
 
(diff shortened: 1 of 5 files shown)


viboes:
What do you think of duplicating the classes timed and not timed and deprecate the timed operations on the not timed classes?

eldiener:
"What do you think of duplicating the classes timed and not timed and deprecate the timed operations on the not timed classes? "

You could do that, but does not the typedef 'duplicate' the classes adequately right now ? I just coded the easiest solution at the present time.

In the future, as C++17 is implemented by compilers and you have already given warning that shared_mutex has been deprecated in favor of share_timed_mutex, you can remove the typedef and actually create a separate non-timed shared_mutex class.

I think it is important now that you add to the documentation that shared_mutex is being deprecated in favor of shared_timed_mutex since shared_mutex's interface will change in the future to a non-timed class. Then if you actually code a non-timed shared_mutex sometime in the future end-users will have been given adequate warning that continued use of the timed operations of shared_mutex will break their code.

viboes:
But 
```
typedef shared_mutex shared_timed_mutex; 
```

will do what you want also, isn't it?

eldiener:
Yes. 

I preferred the other way because in the future you are going to want to keep shared_timed_mutex and change shared_mutex.

PR #60, opened 2015-08-18
Fix to get Thread lib building with Oracle C++

jzmaddock:
See https://svn.boost.org/trac/boost/ticket/11550

This gets Thread building with Oracle C++, there are still test failures, but they're unrelated.


jzmaddock:
Thanks!

PR #62, opened 2015-08-26
Fixed warning C4100 on Windows: Unreferenced Formal Parameter

srouquette:
lk wasn't used here.


viboes:
Already fixed in develop.

PR #63, opened 2015-08-28
remove unused parameter name

sonicbison:
generates warning during compiling


viboes:
Already fixed in develop.
See #62

PR #64, opened 2015-09-11
Fix for future with RTTI off

MarcelRaad:
https://github.com/boostorg/thread/commit/895b4a0ac511f15fdb11964ad4de812695682499 introduced dynamic_casts. This breaks boost::future with RTTI turned off, which worked before. static_casts should be enough here.


viboes:
Yes, we don't need dynamic_cast here. Thanks for the PR.

PR #66, opened 2015-09-19
Avoid deprecated auto_ptr

BenWiederhake:
This fixes the issue described in https://svn.boost.org/trac/boost/ticket/11672 ("Should use unique_ptr, not auto_ptr").

EDIT: I forgot to mention: It passes all regression tests. (`fix-auto-ptr/master` also passes all regression tests.)

This PR is against develop. In case you want a PR against master, see https://github.com/BenWiederhake/thread/tree/fix-auto-ptr/master


viboes:
Thanks

PR #67, opened 2015-09-23
Do not wake up computer from sleep

mauve:
Supplying a REASON_CONTEXT to SetWaitableTimerEx() will wakeup
the computer from sleep. Supplying a nullptr instead disables wake.

Fixes: https://svn.boost.org/trac/boost/ticket/11368

### DIFF
--- a/src/win32/thread.cpp
+++ b/src/win32/thread.cpp
@@ -645,7 +645,6 @@ namespace boost
                     } Detailed;
                 } Reason;
             } REASON_CONTEXT, *PREASON_CONTEXT;
-            static REASON_CONTEXT default_reason_context={0/*POWER_REQUEST_CONTEXT_VERSION*/, 0x00000001/*POWER_REQUEST_CONTEXT_SIMPLE_STRING*/, (LPWSTR)L"generic"};
             typedef BOOL (WINAPI *setwaitabletimerex_t)(HANDLE, const LARGE_INTEGER *, LONG, PTIMERAPCROUTINE, LPVOID, PREASON_CONTEXT, ULONG);
             static inline BOOL WINAPI SetWaitableTimerEx_emulation(HANDLE hTimer, const LARGE_INTEGER *lpDueTime, LONG lPeriod, PTIMERAPCROUTINE pfnCompletionRoutine, LPVOID lpArgToCompletionRoutine, PREASON_CONTEXT WakeContext, ULONG TolerableDelay)
             {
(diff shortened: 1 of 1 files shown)


viboes:
Thanks for the patch. I'm not a Windows expert. We will merge it as soon as Niall has time to check it.

ned14:
This is not a documented valid value for the WakeContext parameter. I would be happy to sign off on this anyway, however neither Wine nor ReactOS fully implement SetWaitableTimerEx() so I cannot tell what the valid values are from them either. The NT kernel docs are similarly not helpful: https://msdn.microsoft.com/en-us/library/windows/hardware/ff553249(v=vs.85).aspx.

Vicente you could try merging this patch and see what happens? As I mentioned, I never saw a problem on my machine with the current code anyway, so it is hard for me to say if this patch is good or bad.

MarcelRaad:
We're successfully using principally the same patch. But the nullptr should probably changed to 0 or NULL in order to support old compilers?

mauve:
we looked at the disassembled source for SetWaitableTimerEx, as it turns out, setting wakecontext to null actually calls SetWaitableTimer :D

mauve:
@MarcelRaad should I switch to NULL instead?

ned14:
> we looked at the disassembled source for SetWaitableTimerEx, as it turns out, setting wakecontext to null actually calls SetWaitableTimer :D

This would suggest that we revert my patch to its previous edition.

It might be worth asking on stackoverflow how to correctly call SetWaitableTimerEx.

Niall

ned14:
Question asked to https://stackoverflow.com/questions/32749829/how-to-call-setwaitabletimerex-correctly

ned14:
I'm thinking that one could pass a NULL WakeContext when wait time exceeds some amount? That would disable PC wake, but keep timer coalescing for shorter values. What's a reasonable amount? Ten seconds?

mauve:
@ned14: I was thinking about that as well. But on the other hand are coalescing timers really useful without wake?

slikts:
I'll just [leave this](https://community.spotify.com/t5/Help-Desktop-Linux-Windows-Web/Windows-Spotify-wakes-computer-from-sleep/m-p/1212180#M140525) here.

viboes:
I can not take this PR as it is for master. I'll propagate this PR to develop and see how it behaves. 

Niall, please, come back to this issue once you have more time.

viboes:
Patch applied to develop branch

https://github.com/boostorg/thread/commit/9ac736a4c4214cdb01caaa84a1121b3846dc87b4

mauve:
@ned14: do you want me to try and revert the following commit or should I just prepare a commit which removes the parts about SetWaitableTimerEx and switch back to SetWaitableTimer? Personally I am unsure if coalescing timers are useful without wakeup (no need to coalesce if already running the CPU at full speed).

    commit 401f69f108271fe2531832a3d85921f019aca421
    Author: Niall Douglas (s [underscore] sourceforge {at} nedprod [dot] com) <spam@nowhere>
    Date:   Wed Feb 4 13:58:11 2015 +0000

        Added coalesced timer support to Windows where that API is available. Tolerable delay is set to the maximum of 5% of interval or 32 ms.

viboes:
Hi Mauve,

I will be very grateful if you go ahead and provide a PR to revert this commit.

Vicente

mauve:
@viboes: Do you mind if I wait with the revert for this PR and do that at the same time I prepare a "proper" fix according to what @ned14 wants me to do? Despite not being beautiful this PR fixes a serious issue.

ned14:
Stackoverflow says the current patch is the right one https://github.com/boostorg/thread/pull/67#issuecomment-144169487

mauve:
@ned14: Perfect thanks, I will prepare a patch to cleanup the remaining unused wakecontext code.

PR #70, opened 2015-10-10
doc: add PROVIDES_EXECUTORS to the config table

mauve:


mauve:
Did you see my email where I was confused that could not find where _PROVIDES_EXECUTORS is not defined in config.hpp? This PR says (probably incorrectly ) that executors are provided when v4 is used. Do you want me to update the table or shall I fix config.hpp?

viboes:
Yes, I read it, but too late :(
Don't worry I will fix the documentation, as I prefer to don't provide it in v4, because the `async()`/`then()` interface changes.

I can include it in V5, but as the interface is not yet stable I prefer to wait.

Other changes in V5 will be to move to non-blocking futures for `then()`and having light Executors (copyable executors).

Vladimir requests associating an Executor to the continuation of a future will be also part of the new version. 

I'm not sure yet if I could have these branches ready for boost1.60.

PR #72, opened 2015-11-03
Remove duplicate conditional check.

ya1gaurav:
thread_info already checked against Null at line no 86, no need to check again.


viboes:
As you see There was a comment `// fixme: should we test this?`.

Some time ago when I removed this check I got in trouble. After checking this seems to work now :)

PR #78, opened 2016-02-19
Fix typo in implementation detail namespace

brycelelbach:
Either I'm missing something or hidden has been misspelled as 'hiden' here.


viboes:
Thanks.

viboes:
Thanks

PR #79, opened 2016-02-27
Fix compilation on Windows

Lastique:
Add a missing include for get_milliseconds_until and namespace qualification.


viboes:
Thnaks.

PR #80, opened 2016-04-16
Fix leak in tls

blastrock:
Hi,

I am experimenting memory leaks with this simple program:

```cpp
#include <boost/thread.hpp>

boost::thread_specific_ptr<int> p;

int main(int argc, char *argv[])
{
  p.reset(new int);
  return 0;
}
```

I compiled boost with gcc 5.3 and the program with clang 3.8, and I linked with boost 1.60 statically. I ran the program through valgrind (clang sanitizer doesn't see all the leaks, I think this is because I didn't compile boost with the sanitizer).

If the code in BOOST_THREAD_PATCH is commented out, I have 6 allocations leaking. If it is enabled, I have only 2. Then with my patch, all boost thread allocations are freed.

May I know why this BOOST_THREAD_PATCH thing exists, and why has it been removed and then restored?

EDIT: I forgot to say that I didn't test this patch on git boost, only on boost 1.60.

PR #81, opened 2016-04-18
fix crash in regression tests because of tls

blastrock:
HI,

Thanks for your answer on #80. This patch seems to fix the regression you found.

Once again, I only tested it on boost 1.61b1 on linux.


viboes:
Many thanks. This works now.

PR #82, opened 2016-04-20
docs: thread_interruption never calls terminate()

akrzemi1:
This note in the docs tries to clarify that exception `thread_interrupted` leaving the thread-main function never triggers a call to `std::terminate`. The previous wording used word 'terminate' in a way that made the semantics ambiguous.


viboes:
Thanks.

PR #86, opened 2016-05-25
Fix Windows version problems

MarcelRaad:
- the target Windows version was hardcoded to Windows 95
  in thread.cpp if not already defined
- the include for BOOST_USE_WINAPI_VERSION was missing

Tested with Visual Studio 2015.


viboes:
I'm unable to check on windows. Hopping this will not introduce a regressions elsewhere.

PR #87, opened 2016-06-20
win32: compile fix for msvc2015

timblechmann:
i had to use this patch to compile against msvc2015

### DIFF
--- a/src/win32/tss_pe.cpp
+++ b/src/win32/tss_pe.cpp
@@ -112,7 +112,7 @@ extern BOOL (WINAPI * const _pDefaultRawDllMainOrig)(HANDLE, DWORD, LPVOID) = NU
 
     //Definitions required by implementation
 
-    #if (_MSC_VER < 1300) // 1300 == VC++ 7.0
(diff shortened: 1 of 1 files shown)


MarcelRaad:
@timblechmann @viboes That's strange as there is no (_MSC_VER > 1900). MSVC 14 (2015) as well as the latest MSVC 15 have _MSC_VER defined as 1900.

Smibu:
@MarcelRaad @timblechmann @viboes This PR appears to break things on VS 2017 RC. For VS 2017 RC, the compiler version is 19.10.24629, so `_MSC_VER` is `1910`. That means `on_tls_prepare` (and the two other functions) will have return type `void`, but the CRT apparently expects a zero value on success.

I encountered this issue when I built my application on VS 2017 RC, and when I ran it, it failed silently. As I debugged it, I noticed `on_tls_prepare` returning an undefined (nonzero) value, and as a result, the program quitted immediately.

I fixed this for myself by reverting the change made in this PR.

PR #89, opened 2016-07-13
fix a bug in upgrade_to_unique_lock<>::operator=()

rick68:
It invokes the copy constructor that already be deleted.


viboes:
Oups!

viboes:
Thanks again.

PR #92, opened 2016-07-16
Fixed compilation with MSVC-8.

Lastique:
Make sure _WIN32_WINNT is defined when windows.h is included. This is
achieved by (a) making sure boost/detail/winapi/config.hpp is included
before any other headers and (b) BOOST_USE_WINDOWS_H is defined so that
the header defines _WIN32_WINNT based on the default target Windows
version. This ensures that all APIs used by the implementation are
available.

Also extracted WIN32_LEAN_AND_MEAN definition to the Jamfile so that it is
consistently defined for all translation units instead of only tss*.cpp.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -126,6 +126,8 @@ project boost/thread
         <toolset>msvc:<cxxflags>/wd4512
         <toolset>msvc:<cxxflags>/wd6246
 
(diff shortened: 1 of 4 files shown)


Lastique:
It would be great if this is merged before the 1.62 release. Is there any problem with the patch?

viboes:
Thanks for working on this issue.

PR #94, opened 2016-07-25
fix bugs in boost::barrier

rick68:
fixes that `type` in `struct boost::result_of<F>` does not name a type.

This is my test code: http://coliru.stacked-crooked.com/a/617f35360b4e1544


viboes:
Already merged. Thanks.

PR #102, opened 2016-10-22
fix boost::synchronized_value<>::load()

rick68:
The data member `boost::synchronized_value<>::value_` is not mutable, so boost::synchronized_value<>::load() could not be const.


viboes:
Thanks. 
Merged.

PR #104, opened 2016-10-24
fix compile failed with boost::user_scheduler

rick68:
The `work` is not in namespace `boost::thread_detail` anymore.


viboes:
Thanks.

PR #106, opened 2016-12-04
Less ambiguous definition for lock(...) functions

apolukhin:
Current definition may confuse library users, they may think that calling `lock(m1,m2)` and `lock(m1, m2,m3, m4)` may result in a deadlock (because they are calling `lock()` for different set of mutexes).


viboes:
Thanks

PR #118, opened 2017-02-28
Loop executor: Don't poll but wait in loop() + test enhanchment.

### LINKED ISSUE #117: loop_executor should block on it's work_queue instead of polling
`loop_executor::loop` is currently:
```c++
    void loop()
    {
      while (!closed())
      {
        schedule_one_or_yield();
      }
      while (try_executing_one())
      {
      }
    }
```

The first loop repeatedly calls `schedule_one_or_yield()` which is simply
```c++
    void schedule_one_or_yield()
    {
        if ( ! try_executing_one())
        {
          this_thread::yield();
        }
    }
```

The current implementation uses unnecessary CPU cycles (esp. if all other threads are idle) by busy waiting. As `concurrent::sync_queue<work> work_queue` already supports a blocking `wait_pull_front` , it would be better to use a (private?) `wait_executing_one` that behaves just like `try_executing_one` but calls `work_queue.wait_pull_front` instead of `try_pull_front`.

IMHO, the current implementation is an outright performance bug.

Oberon00:
Fixes #117.

### DIFF
--- a/include/boost/thread/executors/loop_executor.hpp
+++ b/include/boost/thread/executors/loop_executor.hpp
@@ -16,6 +16,7 @@
 #include <boost/thread/detail/move.hpp>
 #include <boost/thread/concurrent_queues/sync_queue.hpp>
 #include <boost/thread/executors/work.hpp>
+#include <boost/assert.hpp>
 
 #include <boost/config/abi_prefix.hpp>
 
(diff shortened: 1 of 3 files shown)


Oberon00:
I should mention that I only did a `b2 ts_` on my Windows 10 x64 MSVC 14 2015 platform to confirm that the updated test passes for me, but that's all.

Oberon00:
Oh, I just see that the test seems to work just by chance because unless I'm mistaken `~serial_executor` fails do ensure that "The completion of all the closures happen before the completion of the executor destructor. " ([docs](http://www.boost.org/doc/libs/1_63_0/doc/html/thread/synchronization.html#thread.synchronization.executors.ref.serial_executor.destructor)). It seems to just call `close()`.

viboes:
Thanks for the PR

PR #120, opened 2017-03-24
Remove wrong MSVC version check

MarcelRaad:
There is no compiler with `(_MSC_VER > 2000)` and taking the legacy codepath would probably break compilation again when there is. As `(_MSC_VER < 1300)` (that would be MSVC up to 6.0 from 1998) is not supported by Boost anymore anyway and cannot compile Boost.Thread, this codepath can just be removed completely.


MarcelRaad:
Thanks for merging!

PR #122, opened 2017-05-10
Fix sync_bounded_queue<>::size()

vincenthsu:
I create a sync_bounded_queue with capacity 10 and then push 1 item into the queue. The size() functions will get 9, but it should be 1.


viboes:
Thanks a lot for this fix. I really need to add more test :)

PR #126, opened 2017-07-29
Use BOOST_MAY_ALIAS from Boost.Config.

Lastique:
Boost.Config adds support for more compilers that provide the may_alias attribute.


viboes:
Thanks for this PR

PR #128, opened 2017-08-10
Fix for the ticket #13155 - log doesn't build on a system with pthreads

mkartashev:
Changed a call to getpagesize() that is not available since POSIX.1-2001 with
the ::sysconf( _SC_PAGESIZE) call.

The fix was tested by building with "b2 stdlib=sun=stlport" on
- Oracle Linux 7.3,
- Oracle Linux 6.9,
- Solaris 11.3

### DIFF
--- a/include/boost/thread/pthread/thread_data.hpp
+++ b/include/boost/thread/pthread/thread_data.hpp
@@ -35,6 +35,13 @@
 
 #include <boost/config/abi_prefix.hpp>
 
+// Some systems reportedly don't have _SC_PAGESIZE, but do have _SC_PAGE_SIZE
+// instead. The following makes sure that the corresponding sysconf() call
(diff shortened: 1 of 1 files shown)


viboes:
I have tried with 
```
#ifdef BOOST_THREAD_USES_GETPAGESIZE
          std::size_t page_size = getpagesize();
#else
          std::size_t page_size = ::sysconf( _SC_PAGESIZE);
#endif
```

in Macos and it works. I've commit it in develop and will see how it behaves with other platforms.

mkartashev:
Yes, I saw this patch, tested it on Solaris 11.3 and Oracle Linux (6 and 7) and can confirm that it works there. Thanks!

viboes:
Thanks for the information.
I'll close this PR.

PR #131, opened 2017-08-27
Changed signatures of TLS cleanup hooks for consistency with MSVC RTL.

Lastique:
This should fix startup of executables statically linked with Boost.Thread
because of undefined return value of the thread/process startup hooks.

Should fix https://svn.boost.org/trac10/ticket/12730.


viboes:
Thanks a lot.

Issue #130, opened 2017-08-23
windows: Bug in boost::condition_variable on Windows


panyusko:
The bug was made in commit https://github.com/boostorg/thread/commit/ace2b8f89ec398654e9f696d26accbb0b8039bdb (see line 247)
```
// do it here to avoid throwing on the destructor
entry->remove_waiter();
```
Instead should be `entry.remove_waiter();` because of **entry_manager** has overloaded operator -> and the function is called on **basic_cv_list_entry** but not **entry_manager** itself. This results in synchronization issues.

viboes:
I suspect that this was already fixed, isn't it?

viboes:
You are right. Using the same name for the function was an source of errors.

Thanks for catching this.

viboes:
https://github.com/boostorg/thread/commit/a02f0ec577644ca45cdca20e91cd8cd82e6c2017

PR #138, opened 2017-08-28
Fixes/1.65

viboes:
#130 - Bug in boost::condition_variable on Windows.


viboes:
This is already on develop and master

PR #141, opened 2017-08-31
Fix Issues with BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC

austin-beer:
* Fixed https://svn.boost.org/trac10/ticket/12727 (maybe - not tested)
* Fixed the following functions in the pthread implementation to behave correctly when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled and the system time changes:
    * boost::this_thread::sleep_until() with a system clock time
    * boost::thread::try_join_until() with a system clock time
    * boost::condition_variable::wait_until() with a system clock time and predicate
    * boost::condition_variable_any::wait_until() with a system clock time and predicate
    * boost::shared_mutex::try_lock_until() with a system clock time
    * boost::upgrade_mutex::try_lock_until() with a system clock time
    * boost::timed_mutex::try_lock_for()
    * boost::timed_mutex::try_lock_until()
    * boost::recursive_timed_mutex::try_lock_for()
    * boost::recursive_timed_mutex::try_lock_until()
* Fixed the following deprecated functions in the pthread implementation to behave correctly (some would hang) when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled:
    * boost::this_thread::sleep()
    * boost::thread::timed_join()
    * boost::condition_variable::timed_wait()
    * boost::condition_variable_any::timed_wait()
    * boost::shared_mutex::timed_lock()
    * boost::upgrade_mutex::timed_lock()
    * boost::timed_mutex::timed_lock()
    * boost::recursive_timed_mutex::timed_lock()
* Fixed boost::condition_variable_any::timed_wait() in the pthread implementation to correctly handle is_pos_infinity() and is_special() durations.

### DIFF
--- a/include/boost/thread/detail/thread.hpp
+++ b/include/boost/thread/detail/thread.hpp
@@ -157,7 +157,7 @@ namespace boost
     }
 namespace thread_detail {
 #ifdef BOOST_THREAD_USES_CHRONO
-#if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
+#if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC) || defined(BOOST_THREAD_PLATFORM_WIN32)
         typedef chrono::steady_clock internal_clock_t;
 #else
         typedef chrono::system_clock internal_clock_t;
@@ -487,7 +487,7 @@ namespace thread_detail {
         template <class Rep, class Period>
         bool try_join_for(const chrono::duration<Rep, Period>& rel_time)
         {
-          return try_join_until(chrono::steady_clock::now() + rel_time);
+          return try_join_until(thread_detail::internal_clock_t::now() + rel_time);
         }
 #endif
 
@@ -495,21 +495,10 @@ namespace thread_detail {
         bool try_join_until(const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          bool joined= false;
-          do {
-            thread_detail::internal_clock_t::time_point     s_now = thread_detail::internal_clock_t::now();
-            typename Clock::duration   d = ceil<nanoseconds>(t-Clock::now());
-            if (d <= Clock::duration::zero()) return false; // in case the Clock::time_point t is already reached
-            joined = try_join_until(s_now + d);
-          } while (! joined);
-          return true;
-        }
-        template <class Duration>
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& t)
-        {
-          using namespace chrono;
-          typedef time_point<thread_detail::internal_clock_t, nanoseconds> nano_sys_tmpt;
-          return try_join_until(nano_sys_tmpt(ceil<nanoseconds>(t.time_since_epoch())));
+          thread_detail::internal_clock_t::time_point     s_now = thread_detail::internal_clock_t::now();
+          typename Clock::duration   d = ceil<nanoseconds>(t-Clock::now());
+          if (d <= Clock::duration::zero()) return false; // in case the Clock::time_point t is already reached
+          return try_join_until(s_now + d);
         }
 #endif
 #if defined(BOOST_THREAD_PLATFORM_WIN32)
@@ -523,9 +512,10 @@ namespace thread_detail {
         //}
 
 #ifdef BOOST_THREAD_USES_CHRONO
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, chrono::nanoseconds>& tp)
+        template <class Duration>
+        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
         {
-          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-chrono::system_clock::now());
+          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-thread_detail::internal_clock_t::now());
           return do_try_join_until(rel_time.count());
         }
 #endif
@@ -539,17 +529,14 @@ namespace thread_detail {
 #if defined BOOST_THREAD_USES_DATETIME
         bool timed_join(const system_time& abs_time)
         {
-          struct timespec const ts=detail::to_timespec(abs_time);
-          return do_try_join_until(ts);
+          return do_try_join_until(detail::timespec_to_internal_clock(abs_time));
         }
 #endif
 #ifdef BOOST_THREAD_USES_CHRONO
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, chrono::nanoseconds>& tp)
+        template <class Duration>
+        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
         {
-          using namespace chrono;
-          nanoseconds d = tp.time_since_epoch();
-          timespec ts = boost::detail::to_timespec(d);
-          return do_try_join_until(ts);
+          return do_try_join_until(boost::detail::to_timespec(tp.time_since_epoch()));
         }
 #endif
 
@@ -560,7 +547,7 @@ namespace thread_detail {
         template<typename TimeDuration>
         inline bool timed_join(TimeDuration const& rel_time)
         {
-            return timed_join(get_system_time()+rel_time);
+            return do_try_join_until(detail::timespec_plus_internal_clock(rel_time));
         }
 #endif
         void detach();
--- a/include/boost/thread/detail/thread.hpp
+++ b/include/boost/thread/detail/thread.hpp
@@ -155,15 +155,6 @@ namespace boost
         };
 #endif
     }
-namespace thread_detail {
-#ifdef BOOST_THREAD_USES_CHRONO
-#if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC) || defined(BOOST_THREAD_PLATFORM_WIN32)
-        typedef chrono::steady_clock internal_clock_t;
-#else
-        typedef chrono::system_clock internal_clock_t;
-#endif
-#endif
-}
     class BOOST_THREAD_DECL thread
     {
     public:
@@ -487,19 +478,9 @@ namespace thread_detail {
         template <class Rep, class Period>
         bool try_join_for(const chrono::duration<Rep, Period>& rel_time)
         {
-          return try_join_until(thread_detail::internal_clock_t::now() + rel_time);
+          return do_try_join_until(boost::detail::to_abs_internal_timespec(rel_time));
         }
 #endif
-
-        template <class Clock, class Duration>
-        bool try_join_until(const chrono::time_point<Clock, Duration>& t)
-        {
-          using namespace chrono;
-          thread_detail::internal_clock_t::time_point     s_now = thread_detail::internal_clock_t::now();
-          typename Clock::duration   d = ceil<nanoseconds>(t-Clock::now());
-          if (d <= Clock::duration::zero()) return false; // in case the Clock::time_point t is already reached
-          return try_join_until(s_now + d);
-        }
 #endif
 #if defined(BOOST_THREAD_PLATFORM_WIN32)
     private:
@@ -512,10 +493,10 @@ namespace thread_detail {
         //}
 
 #ifdef BOOST_THREAD_USES_CHRONO
-        template <class Duration>
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
+        template <class Clock, class Duration>
+        bool try_join_until(const chrono::time_point<Clock, Duration>& tp)
         {
-          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-thread_detail::internal_clock_t::now());
+          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-Clock::now());
           return do_try_join_until(rel_time.count());
         }
 #endif
@@ -529,14 +510,14 @@ namespace thread_detail {
 #if defined BOOST_THREAD_USES_DATETIME
         bool timed_join(const system_time& abs_time)
         {
-          return do_try_join_until(detail::timespec_to_internal_clock(abs_time));
+          return do_try_join_until(boost::detail::to_abs_internal_timespec(abs_time));
         }
 #endif
 #ifdef BOOST_THREAD_USES_CHRONO
-        template <class Duration>
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
+        template <class Clock, class Duration>
+        bool try_join_until(const chrono::time_point<Clock, Duration>& tp)
         {
-          return do_try_join_until(boost::detail::to_timespec(tp.time_since_epoch()));
+          return do_try_join_until(boost::detail::to_abs_internal_timespec(tp));
         }
 #endif
 
@@ -547,7 +528,7 @@ namespace thread_detail {
         template<typename TimeDuration>
         inline bool timed_join(TimeDuration const& rel_time)
         {
-            return do_try_join_until(detail::timespec_plus_internal_clock(rel_time));
+            return timed_join(get_system_time()+rel_time);
         }
 #endif
         void detach();
--- a/include/boost/thread/pthread/condition_variable.hpp
+++ b/include/boost/thread/pthread/condition_variable.hpp
@@ -169,14 +169,6 @@ namespace boost
         pthread_mutex_t internal_mutex;
         pthread_cond_t cond;
 
-#ifdef BOOST_THREAD_USES_CHRONO
-#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-        typedef chrono::steady_clock internal_clock_t;
-#else
-        typedef chrono::system_clock internal_clock_t;
-#endif
-#endif
-
     public:
         BOOST_THREAD_NO_COPYABLE(condition_variable_any)
         condition_variable_any()
@@ -186,11 +178,11 @@ namespace boost
             {
                 boost::throw_exception(thread_resource_error(res, "boost::condition_variable_any::condition_variable_any() failed in pthread_mutex_init"));
             }
-            int const res2 = detail::monotonic_pthread_cond_init(cond);
+            int const res2=boost::detail::cond_init(cond);
             if(res2)
             {
                 BOOST_VERIFY(!pthread_mutex_destroy(&internal_mutex));
-                boost::throw_exception(thread_resource_error(res2, "boost::condition_variable_any::condition_variable_any() failed in detail::monotonic_pthread_cond_init"));
+                boost::throw_exception(thread_resource_error(res2, "boost::condition_variable_any::condition_variable_any() failed in boost::detail::cond_init"));
             }
         }
         ~condition_variable_any()
@@ -234,7 +226,7 @@ namespace boost
         template<typename lock_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time)
         {
-            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time));
+            return do_wait_until(m,boost::detail::to_abs_internal_timespec(abs_time));
         }
         template<typename lock_type>
         bool timed_wait(lock_type& m,xtime const& abs_time)
@@ -254,13 +246,13 @@ namespace boost
             {
                 return true;
             }
-            return do_wait_until(m,detail::timespec_plus_internal_clock(wait_duration));
+            return do_wait_until(m,boost::detail::to_abs_internal_timespec(wait_duration));
         }
 
         template<typename lock_type,typename predicate_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time, predicate_type pred)
         {
-            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time),pred);
+            return do_wait_until(m,boost::detail::to_abs_internal_timespec(abs_time),pred);
         }
 
         template<typename lock_type,typename predicate_type>
@@ -284,31 +276,19 @@ namespace boost
             {
                 return pred();
             }
-            return do_wait_until(m,detail::timespec_plus_internal_clock(wait_duration),pred);
+            return do_wait_until(m,boost::detail::to_abs_internal_timespec(wait_duration),pred);
         }
 #endif // defined BOOST_THREAD_USES_DATETIME
 
 #ifdef BOOST_THREAD_USES_CHRONO
-        template <class lock_type,class Duration>
-        cv_status
-        wait_until(
-                lock_type& lock,
-                const chrono::time_point<internal_clock_t, Duration>& t)
-        {
-          return do_wait_until(lock, boost::detail::to_timespec(t.time_since_epoch()))
-            ? cv_status::no_timeout : cv_status::timeout;
-        }
-
         template <class lock_type, class Clock, class Duration>
         cv_status
         wait_until(
                 lock_type& lock,
                 const chrono::time_point<Clock, Duration>& t)
         {
-          using namespace chrono;
-          internal_clock_t::time_point s_now = internal_clock_t::now();
-          typename Clock::time_point  c_now = Clock::now();
-          return wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
+          return do_wait_until(lock, boost::detail::to_abs_internal_timespec(t))
+            ? cv_status::no_timeout : cv_status::timeout;
         }
 
         template <class lock_type, class Rep, class Period>
@@ -317,17 +297,8 @@ namespace boost
                 lock_type& lock,
                 const chrono::duration<Rep, Period>& d)
         {
-          return wait_until(lock, internal_clock_t::now() + d);
-        }
-
-        template <class lock_type, class Duration, class Predicate>
-        bool
-        wait_until(
-                lock_type& lock,
-                const chrono::time_point<internal_clock_t, Duration>& t,
-                Predicate pred)
-        {
-            return do_wait_until(lock, boost::detail::to_timespec(t.time_since_epoch()), boost::move(pred));
+          return do_wait_until(lock, boost::detail::to_abs_internal_timespec(d))
+            ? cv_status::no_timeout : cv_status::timeout;
         }
 
         template <class lock_type, class Clock, class Duration, class Predicate>
@@ -337,10 +308,7 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t,
                 Predicate pred)
         {
-            using namespace chrono;
-            internal_clock_t::time_point s_now = internal_clock_t::now();
-            typename Clock::time_point  c_now = Clock::now();
-            return wait_until(lock, s_now + ceil<nanoseconds>(t - c_now), boost::move(pred));
+          return do_wait_until(lock, boost::detail::to_abs_internal_timespec(t), boost::move(pred));
         }
 
         template <class lock_type, class Rep, class Period, class Predicate>
@@ -350,7 +318,7 @@ namespace boost
                 const chrono::duration<Rep, Period>& d,
                 Predicate pred)
         {
-          return wait_until(lock, internal_clock_t::now() + d, boost::move(pred));
+          return do_wait_until(lock, boost::detail::to_abs_internal_timespec(d), boost::move(pred));
         }
 #endif // defined BOOST_THREAD_USES_CHRONO
 
--- a/include/boost/thread/pthread/condition_variable.hpp
+++ b/include/boost/thread/pthread/condition_variable.hpp
@@ -134,6 +134,20 @@ namespace boost
         return true;
     }
 
+    template<typename predicate_type>
+    bool condition_variable::do_wait_until(
+         unique_lock<mutex>& m,
+         struct timespec const &timeout,
+         predicate_type pred)
+    {
+        while (!pred())
+        {
+            if (!do_wait_until(m, timeout))
+                return pred();
+        }
+        return true;
+    }
+
     inline void condition_variable::notify_one() BOOST_NOEXCEPT
     {
 #if defined BOOST_THREAD_PROVIDES_INTERRUPTIONS
@@ -154,6 +168,14 @@ namespace boost
     {
         pthread_mutex_t internal_mutex;
         pthread_cond_t cond;
+
+#ifdef BOOST_THREAD_USES_CHRONO
+#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+        typedef chrono::steady_clock internal_clock_t;
+#else
+        typedef chrono::system_clock internal_clock_t;
+#endif
+#endif
 
     public:
         BOOST_THREAD_NO_COPYABLE(condition_variable_any)
@@ -212,8 +234,7 @@ namespace boost
         template<typename lock_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time)
         {
-            struct timespec const timeout=detail::to_timespec(abs_time);
-            return do_wait_until(m, timeout);
+            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time));
         }
         template<typename lock_type>
         bool timed_wait(lock_type& m,xtime const& abs_time)
@@ -224,18 +245,22 @@ namespace boost
         template<typename lock_type,typename duration_type>
         bool timed_wait(lock_type& m,duration_type const& wait_duration)
         {
-            return timed_wait(m,get_system_time()+wait_duration);
+            if (wait_duration.is_pos_infinity())
+            {
+                wait(m); // or do_wait(m,detail::timeout::sentinel());
+                return true;
+            }
+            if (wait_duration.is_special())
+            {
+                return true;
+            }
+            return do_wait_until(m,detail::timespec_plus_internal_clock(wait_duration));
         }
 
         template<typename lock_type,typename predicate_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time, predicate_type pred)
         {
-            while (!pred())
-            {
-                if(!timed_wait(m, abs_time))
-                    return pred();
-            }
-            return true;
+            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time),pred);
         }
 
         template<typename lock_type,typename predicate_type>
@@ -247,24 +272,31 @@ namespace boost
         template<typename lock_type,typename duration_type,typename predicate_type>
         bool timed_wait(lock_type& m,duration_type const& wait_duration,predicate_type pred)
         {
-            return timed_wait(m,get_system_time()+wait_duration,pred);
+            if (wait_duration.is_pos_infinity())
+            {
+                while (!pred())
+                {
+                  wait(m); // or do_wait(m,detail::timeout::sentinel());
+                }
+                return true;
+            }
+            if (wait_duration.is_special())
+            {
+                return pred();
+            }
+            return do_wait_until(m,detail::timespec_plus_internal_clock(wait_duration),pred);
         }
-#endif
-#ifndef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif // defined BOOST_THREAD_USES_DATETIME
 
 #ifdef BOOST_THREAD_USES_CHRONO
         template <class lock_type,class Duration>
         cv_status
         wait_until(
                 lock_type& lock,
-                const chrono::time_point<chrono::system_clock, Duration>& t)
+                const chrono::time_point<internal_clock_t, Duration>& t)
         {
-          using namespace chrono;
-          typedef time_point<system_clock, nanoseconds> nano_sys_tmpt;
-          wait_until(lock,
-                        nano_sys_tmpt(ceil<nanoseconds>(t.time_since_epoch())));
-          return system_clock::now() < t ? cv_status::no_timeout :
-                                             cv_status::timeout;
+          return do_wait_until(lock, boost::detail::to_timespec(t.time_since_epoch()))
+            ? cv_status::no_timeout : cv_status::timeout;
         }
 
         template <class lock_type, class Clock, class Duration>
@@ -274,10 +306,9 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          system_clock::time_point     s_now = system_clock::now();
+          internal_clock_t::time_point s_now = internal_clock_t::now();
           typename Clock::time_point  c_now = Clock::now();
-          wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
-          return Clock::now() < t ? cv_status::no_timeout : cv_status::timeout;
+          return wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
         }
 
         template <class lock_type, class Rep, class Period>
@@ -286,86 +317,19 @@ namespace boost
                 lock_type& lock,
                 const chrono::duration<Rep, Period>& d)
         {
-          using namespace chrono;
-          system_clock::time_point s_now = system_clock::now();
-          steady_clock::time_point c_now = steady_clock::now();
-          wait_until(lock, s_now + ceil<nanoseconds>(d));
-          return steady_clock::now() - c_now < d ? cv_status::no_timeout :
-                                                   cv_status::timeout;
-
+          return wait_until(lock, internal_clock_t::now() + d);
         }
 
-        template <class lock_type>
-        cv_status wait_until(
-            lock_type& lk,
-            chrono::time_point<chrono::system_clock, chrono::nanoseconds> tp)
-        {
-            using namespace chrono;
-            nanoseconds d = tp.time_since_epoch();
-            timespec ts = boost::detail::to_timespec(d);
-            if (do_wait_until(lk, ts)) return cv_status::no_timeout;
-            else return cv_status::timeout;
-        }
-#endif
-#else // defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-#ifdef BOOST_THREAD_USES_CHRONO
-
-        template <class lock_type, class Duration>
-        cv_status
+        template <class lock_type, class Duration, class Predicate>
+        bool
         wait_until(
-            lock_type& lock,
-            const chrono::time_point<chrono::steady_clock, Duration>& t)
+                lock_type& lock,
+                const chrono::time_point<internal_clock_t, Duration>& t,
+                Predicate pred)
         {
-            using namespace chrono;
-            typedef time_point<steady_clock, nanoseconds> nano_sys_tmpt;
-            wait_until(lock,
-                        nano_sys_tmpt(ceil<nanoseconds>(t.time_since_epoch())));
-            return steady_clock::now() < t ? cv_status::no_timeout :
-                                             cv_status::timeout;
+            return do_wait_until(lock, boost::detail::to_timespec(t.time_since_epoch()), boost::move(pred));
         }
 
-        template <class lock_type, class Clock, class Duration>
-        cv_status
-        wait_until(
-            lock_type& lock,
-            const chrono::time_point<Clock, Duration>& t)
-        {
-            using namespace chrono;
-            steady_clock::time_point     s_now = steady_clock::now();
-            typename Clock::time_point  c_now = Clock::now();
-            wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
-            return Clock::now() < t ? cv_status::no_timeout : cv_status::timeout;
-        }
-
-        template <class lock_type, class Rep, class Period>
-        cv_status
-        wait_for(
-            lock_type& lock,
-            const chrono::duration<Rep, Period>& d)
-        {
-            using namespace chrono;
-            steady_clock::time_point c_now = steady_clock::now();
-            wait_until(lock, c_now + ceil<nanoseconds>(d));
-            return steady_clock::now() - c_now < d ? cv_status::no_timeout :
-                                                   cv_status::timeout;
-        }
-
-        template <class lock_type>
-        inline cv_status wait_until(
-            lock_type& lock,
-            chrono::time_point<chrono::steady_clock, chrono::nanoseconds> tp)
-        {
-            using namespace chrono;
-            nanoseconds d = tp.time_since_epoch();
-            timespec ts = boost::detail::to_timespec(d);
-            if (do_wait_until(lock, ts)) return cv_status::no_timeout;
-            else return cv_status::timeout;
-        }
-
-#endif
-#endif // defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-
-#ifdef BOOST_THREAD_USES_CHRONO
         template <class lock_type, class Clock, class Duration, class Predicate>
         bool
         wait_until(
@@ -373,12 +337,10 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t,
                 Predicate pred)
         {
-            while (!pred())
-            {
-                if (wait_until(lock, t) == cv_status::timeout)
-                    return pred();
-            }
-            return true;
+            using namespace chrono;
+            internal_clock_t::time_point s_now = internal_clock_t::now();
+            typename Clock::time_point  c_now = Clock::now();
+            return wait_until(lock, s_now + ceil<nanoseconds>(t - c_now), boost::move(pred));
         }
 
         template <class lock_type, class Rep, class Period, class Predicate>
@@ -388,9 +350,9 @@ namespace boost
                 const chrono::duration<Rep, Period>& d,
                 Predicate pred)
         {
-          return wait_until(lock, chrono::steady_clock::now() + d, boost::move(pred));
+          return wait_until(lock, internal_clock_t::now() + d, boost::move(pred));
         }
-#endif
+#endif // defined BOOST_THREAD_USES_CHRONO
 
         void notify_one() BOOST_NOEXCEPT
         {
@@ -435,6 +397,20 @@ namespace boost
               boost::throw_exception(condition_error(res, "boost::condition_variable_any::do_wait_until() failed in pthread_cond_timedwait"));
           }
           return true;
+        }
+
+        template <class lock_type, typename predicate_type>
+        bool do_wait_until(
+          lock_type& m,
+          struct timespec const &timeout,
+          predicate_type pred)
+        {
+            while (!pred())
+            {
+                if (!do_wait_until(m, timeout))
+                    return pred();
+            }
+            return true;
         }
     };
 
--- a/include/boost/thread/pthread/mutex.hpp
+++ b/include/boost/thread/pthread/mutex.hpp
@@ -36,12 +36,13 @@
 #endif
 #endif
 
-// CLOCK_MONOTONIC only works with pthread_cond_timedwait(), not with pthread_mutex_timedlock()
-#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-#undef BOOST_PTHREAD_HAS_TIMEDLOCK
-#endif
 
 #include <boost/config/abi_prefix.hpp>
+
+// CLOCK_MONOTONIC only works with pthread_cond_timedwait(), not with pthread_mutex_timedlock()
+#if defined(BOOST_PTHREAD_HAS_TIMEDLOCK) && !defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
+#define BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
+#endif
 
 #ifndef BOOST_THREAD_HAS_NO_EINTR_BUG
 #define BOOST_THREAD_HAS_EINTR_BUG
@@ -171,19 +172,10 @@ namespace boost
     {
     private:
         pthread_mutex_t m;
-#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK
+#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
         pthread_cond_t cond;
         bool is_locked;
 #endif
-
-#ifdef BOOST_THREAD_USES_CHRONO
-#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-        typedef chrono::steady_clock internal_clock_t;
-#else
-        typedef chrono::system_clock internal_clock_t;
-#endif
-#endif
-
     public:
         BOOST_THREAD_NO_COPYABLE(timed_mutex)
         timed_mutex()
@@ -193,32 +185,21 @@ namespace boost
             {
                 boost::throw_exception(thread_resource_error(res, "boost:: timed_mutex constructor failed in pthread_mutex_init"));
             }
-#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK
-#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
-            pthread_condattr_t attr;
-            int res2=pthread_condattr_init(&attr);
-            if (!res2)
-            {
-                pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
-                res2=pthread_cond_init(&cond,&attr);
-                pthread_condattr_destroy(&attr);
-            }
-#else
-            int const res2=pthread_cond_init(&cond,NULL);
-#endif
+#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
+            int const res2=boost::detail::cond_init(cond);
             if(res2)
             {
                 BOOST_VERIFY(!posix::pthread_mutex_destroy(&m));
                 //BOOST_VERIFY(!pthread_mutex_destroy(&m));
-                boost::throw_exception(thread_resource_error(res2, "boost:: timed_mutex constructor failed in pthread_cond_init"));
+                boost::throw_exception(thread_resource_error(res2, "boost:: timed_mutex constructor failed in boost::detail::cond_init"));
             }
             is_locked=false;
-#endif // !defined BOOST_PTHREAD_HAS_TIMEDLOCK
+#endif // !defined BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
         }
         ~timed_mutex()
         {
             BOOST_VERIFY(!posix::pthread_mutex_destroy(&m));
-#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK
+#ifndef BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
             BOOST_VERIFY(!pthread_cond_destroy(&cond));
 #endif
         }
@@ -227,14 +208,14 @@ namespace boost
         template<typename TimeDuration>
         bool timed_lock(TimeDuration const & relative_time)
         {
-            return do_try_lock_until(boost::detail::timespec_plus_internal_clock(relative_time));
+            return do_try_lock_until(boost::detail::to_abs_internal_timespec(relative_time));
         }
         bool timed_lock(boost::xtime const & absolute_time)
         {
             return timed_lock(system_time(absolute_time));
         }
 #endif
-#ifdef BOOST_PTHREAD_HAS_TIMEDLOCK
+#ifdef BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
         void lock()
         {
             int res = posix::pthread_mutex_lock(&m);
@@ -280,7 +261,7 @@ namespace boost
         }
     public:
 
-#else // !defined BOOST_PTHREAD_HAS_TIMEDLOCK
+#else // !defined BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
         void lock()
         {
             boost::pthread::pthread_mutex_scoped_lock const local_lock(&m);
@@ -326,32 +307,24 @@ namespace boost
             return true;
         }
     public:
-#endif // !defined BOOST_PTHREAD_HAS_TIMEDLOCK
+#endif // !defined BOOST_PTHREAD_HAS_TIMEDLOCK_LOCAL
 
 #if defined BOOST_THREAD_USES_DATETIME
         bool timed_lock(system_time const & abs_time)
         {
-            return do_try_lock_until(boost::detail::timespec_to_internal_clock(abs_time));
+            return do_try_lock_until(boost::detail::to_abs_internal_timespec(abs_time));
         }
 #endif
 #ifdef BOOST_THREAD_USES_CHRONO
         template <class Rep, class Period>
         bool try_lock_for(const chrono::duration<Rep, Period>& rel_time)
         {
-          return try_lock_until(internal_clock_t::now() + rel_time);
+          return do_try_lock_until(boost::detail::to_abs_internal_timespec(rel_time));
         }
         template <class Clock, class Duration>
         bool try_lock_until(const chrono::time_point<Clock, Duration>& t)
         {
-          using namespace chrono;
-          internal_clock_t::time_point s_now = internal_clock_t::now();
-          typename Clock::time_point  c_now = Clock::now();
-          return try_lock_until(s_now + ceil<nanoseconds>(t - c_now));
-        }
-        template <class Duration>
-        bool try_lock_until(const chrono::time_point<internal_clock_t, Duration>& t)
-        {
-          return do_try_lock_until(boost::detail::to_timespec(t.time_since_epoch()));
+          return do_try_lock_until(boost::detail::to_abs_internal_timespec(t));
         }
 #endif // defined BOOST_THREAD_USES_CHRONO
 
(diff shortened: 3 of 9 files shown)


brainwave64:
I realize these changes appear fairly extensive, but I believe they are warranted in order to ensure that BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC behaves consistently and reliably regardless of which time-related functions users call.

My hope is that, once these changes have been well-tested, BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC can be replaced by CLOCK_MONOTONIC so that all users can benefit (by default) from the monotonic clock whenever it is available.

Here is the executable that I've been using to test these updates. It executes each function and then changes the system time while the function is executing and then measures how long the function took to actually execute.

[test_boost_sleep.cpp.txt](https://github.com/boostorg/thread/files/1268037/test_boost_sleep.cpp.txt)

Austin Beer

viboes:
Hi, thanks for your interest and the hard work you have done.
I recognize that I have had a lot of trouble with the time since I started to try to improve it with monotonic clocks.

I will not accept your PR as such as it introduce too much changes. 

I would prefer to do it in small steps if your prefer.

Once of my major troubles have been to use the same timespec.
I wonder if it is not better to wrap timespec in two classes: 

* mono_timespec
* real_timespec

and define the needed operations on these classes, now, plus, minus, ...

and then have conversions from  Chrono types and DateTime types to timespec types

* to_mono_timespec
* to_real_timespec

mono_timespec will not exist in systems that don't have MONOTONIC clocks.

The other possible error I did is to convert between monotonic and real times. As we can not convert between steady clocks and system clocks, we can not convert between monotonic and real timespec. 

BTW, is the time provided by DateTime monotonic (steady) or real(system).
If it provides system time we cannot convert it to monotonic.

Once this is clear we could start to build on top of these clocks.

Please, let me know what do you think and thanks for your interest.

viboes:
I'll create a branch to do all the changes needed to work with monotonic timespec, so that we can check on various platforms before going to trunk.

viboes:
Wondering if in addition I'm not mixing some times timespec that convey relative or absolute times :(

So we will need 
* absolute_mono_timespec 
* absolute_real_timespec 
* relative_timespec

brainwave64:
Hi Vicente,

Thanks for taking the time to consider this pull request. I did wonder if the extent of the changes would be an issue. Unfortunately in order to fully support monotonic time I believe all of the changes will ultimately need to be implemented. However, these changes could certainly be broken up to a series of smaller commits so that each commit is easier to review and manage.

I agree that it would be helpful to wrap the timespecs in classes so that it's clear what the timespec is representing. In addition to the three you already mention, we probably also need an "absolute_internal_timespec" class that is monotonic if the monotonic clock is enabled and realtime if it is not. Functions like condition_variable::do_wait_until() would need to be passed this type because the type of timespec these functions need depends on whether or not the monotonic clock is enabled.

I think it is good to be able to convert between monotonic and realtime clocks, provided it is clear that that's what we are doing and that the conversion happens only once, when the user first calls a function and passes a time to us.

I made the assumption that the DateTime times were always system/realtime, since as far as I can tell the Boost.DateTime library doesn't have a concept of a monotonic clock.

I like the idea of having a branch to work with and test on various platforms. Please let me know how I can help.

Austin

include/boost/thread/detail/thread.hpp
```diff
@@ -539,17 +529,14 @@ namespace thread_detail {
 #if defined BOOST_THREAD_USES_DATETIME
         bool timed_join(const system_time& abs_time)
         {
-          struct timespec const ts=detail::to_timespec(abs_time);
-          return do_try_join_until(ts);
+          return do_try_join_until(detail::timespec_to_internal_clock(abs_time));
```
viboes: I don't think we can convert system_time to internal_clock when internal_clock is steady_clock, as system_time is based on system_clock.

In this case, we should iterate as done in line 495

brainwave64: I believe the timespec_to_internal_clock function handles this case. When the monotonic clock is enabled, it takes the difference between the abs_time and the current system time, and then adds that to the current steady clock, and that's what it returns. That ensures that we wait for the correct amount of time, even if the system clock jumps while we're waiting.

include/boost/thread/detail/thread.hpp
```diff
@@ -560,7 +547,7 @@ namespace thread_detail {
         template<typename TimeDuration>
         inline bool timed_join(TimeDuration const& rel_time)
         {
-            return timed_join(get_system_time()+rel_time);
+            return do_try_join_until(detail::timespec_plus_internal_clock(rel_time));
```
viboes: This has a sense only on Pthread. Windows has no timespec.

include/boost/thread/detail/thread.hpp
```diff
@@ -539,17 +529,14 @@ namespace thread_detail {
...
 #ifdef BOOST_THREAD_USES_CHRONO
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, chrono::nanoseconds>& tp)
+        template <class Duration>
+        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
         {
-          using namespace chrono;
-          nanoseconds d = tp.time_since_epoch();
-          timespec ts = boost::detail::to_timespec(d);
-          return do_try_join_until(ts);
+          return do_try_join_until(boost::detail::to_timespec(tp.time_since_epoch()));
```
viboes: This has a sense only on Pthread. Windows has no timespec.

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -155,6 +169,14 @@ namespace boost
         pthread_mutex_t internal_mutex;
         pthread_cond_t cond;
 
+#ifdef BOOST_THREAD_USES_CHRONO
+#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
```
viboes: This type should be declared only once. What about moving it to a specific file?

brainwave64: That's a good idea.

include/boost/thread/detail/thread.hpp
brainwave64: Agreed. This particular instance is in an #else section that's only enabled when BOOST_THREAD_PLATFORM_WIN32 is not defined.

include/boost/thread/detail/thread.hpp
brainwave64: Agreed. This particular instance is in an #else section that's only enabled when BOOST_THREAD_PLATFORM_WIN32 is not defined.

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -224,18 +245,22 @@ namespace boost
...
         template<typename lock_type,typename predicate_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time, predicate_type pred)
         {
-            while (!pred())
-            {
-                if(!timed_wait(m, abs_time))
-                    return pred();
-            }
-            return true;
+            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time),pred);
```
viboes: system_time and internal_clock can not be converted when the clock is steady

brainwave64: I believe the timespec_to_internal_clock function handles this case. When the monotonic clock is enabled, it takes the difference between the abs_time and the current system time, and then adds that to the current steady clock, and that's what it returns. That ensures that we wait for the correct amount of time, even if the system clock jumps while we're waiting.

include/boost/thread/pthread/mutex.hpp
```diff
@@ -34,6 +36,10 @@
 #endif
 #endif
 
+// CLOCK_MONOTONIC only works with pthread_cond_timedwait(), not with pthread_mutex_timedlock()
+#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#undef BOOST_PTHREAD_HAS_TIMEDLOCK
```
viboes: We can not undef this here. This is a global define. If Boost.Thread can not use pthread_mutex_timedlock() when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is defined, this should stay internal to Boost.Thread

brainwave64: Ah, good to know. I'll have to fix that.

include/boost/thread/pthread/mutex.hpp
```diff
@@ -179,15 +194,26 @@ namespace boost
                 boost::throw_exception(thread_resource_error(res, "boost:: timed_mutex constructor failed in pthread_mutex_init"));
             }
 #ifndef BOOST_PTHREAD_HAS_TIMEDLOCK
+#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
```
viboes: As this would appear several times, I believe it will be word defining an internal thread_detail::pthread_cond_init function that do this

brainwave64: Good idea.

include/boost/thread/v2/thread.hpp
```diff
@@ -69,7 +69,11 @@ namespace boost
       using namespace chrono;
       if (d > duration<Rep, Period>::zero())
       {
-        steady_clock::time_point c_timeout = steady_clock::now() + ceil<nanoseconds>(d);
+#ifdef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
```
viboes: Why not use internal_clock here?

brainwave64: Didn't think to, but that's a good idea.

include/boost/thread/detail/thread.hpp
```diff
@@ -487,29 +487,18 @@ namespace thread_detail {
         template <class Rep, class Period>
         bool try_join_for(const chrono::duration<Rep, Period>& rel_time)
         {
-          return try_join_until(chrono::steady_clock::now() + rel_time);
+          return try_join_until(thread_detail::internal_clock_t::now() + rel_time);
```
viboes: Agreed.

include/boost/thread/detail/thread.hpp
```diff
@@ -523,9 +512,10 @@ namespace thread_detail {
         //}
 
 #ifdef BOOST_THREAD_USES_CHRONO
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, chrono::nanoseconds>& tp)
+        template <class Duration>
+        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
         {
-          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-chrono::system_clock::now());
+          chrono::milliseconds rel_time= chrono::ceil<chrono::milliseconds>(tp-thread_detail::internal_clock_t::now());
```
viboes: Agreed

include/boost/thread/detail/thread.hpp
```diff
@@ -487,29 +487,18 @@ namespace thread_detail {
...
-          } while (! joined);
-          return true;
-        }
-        template <class Duration>
-        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& t)
-        {
-          using namespace chrono;
-          typedef time_point<thread_detail::internal_clock_t, nanoseconds> nano_sys_tmpt;
-          return try_join_until(nano_sys_tmpt(ceil<nanoseconds>(t.time_since_epoch())));
+          thread_detail::internal_clock_t::time_point     s_now = thread_detail::internal_clock_t::now();
```
viboes: I don't agree here. You are converting from Clock to internal_clock.

include/boost/thread/pthread/condition_variable.hpp
viboes: What about thread/detail/internal_clock.hpp?

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -212,8 +234,7 @@ namespace boost
         template<typename lock_type>
         bool timed_wait(lock_type& m,boost::system_time const& abs_time)
         {
-            struct timespec const timeout=detail::to_timespec(abs_time);
-            return do_wait_until(m, timeout);
+            return do_wait_until(m,detail::timespec_to_internal_clock(abs_time));
```
viboes: I don't agree here. system_time and internal_clock are not convertible

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -224,18 +245,22 @@ namespace boost
         template<typename lock_type,typename duration_type>
         bool timed_wait(lock_type& m,duration_type const& wait_duration)
         {
-            return timed_wait(m,get_system_time()+wait_duration);
+            if (wait_duration.is_pos_infinity())
```
viboes: Agreed.

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -247,24 +272,31 @@ namespace boost
         template<typename lock_type,typename duration_type,typename predicate_type>
         bool timed_wait(lock_type& m,duration_type const& wait_duration,predicate_type pred)
         {
-            return timed_wait(m,get_system_time()+wait_duration,pred);
+            if (wait_duration.is_pos_infinity())
```
viboes: Agreed

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -247,24 +272,31 @@ namespace boost
...
-#ifndef BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif // defined BOOST_THREAD_USES_DATETIME
 
 #ifdef BOOST_THREAD_USES_CHRONO
         template <class lock_type,class Duration>
         cv_status
         wait_until(
                 lock_type& lock,
-                const chrono::time_point<chrono::system_clock, Duration>& t)
+                const chrono::time_point<internal_clock_t, Duration>& t)
```
viboes: Agreed

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -274,10 +306,9 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          system_clock::time_point     s_now = system_clock::now();
+          internal_clock_t::time_point s_now = internal_clock_t::now();
```
viboes: Agreed

include/boost/thread/detail/thread.hpp
brainwave64: I guess I'm not sure where the issue is. The code is taking the difference between t and Clock::now(), adding that difference to internal_clock_t::now(), and then passing the result to try_join_until(). Later on in this file, try_join_until() converts the internal clock time to a timespec and passes it to the pthreads implementation of do_try_join_until(), which expects an internal clock time.

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -286,99 +317,30 @@ namespace boost
...
                 const chrono::duration<Rep, Period>& d)
         {
-          using namespace chrono;
-          system_clock::time_point s_now = system_clock::now();
-          steady_clock::time_point c_now = steady_clock::now();
-          wait_until(lock, s_now + ceil<nanoseconds>(d));
-          return steady_clock::now() - c_now < d ? cv_status::no_timeout :
-                                                   cv_status::timeout;
-
+          return wait_until(lock, internal_clock_t::now() + d);
```
viboes: Agreed

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -274,10 +306,9 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          system_clock::time_point     s_now = system_clock::now();
+          internal_clock_t::time_point s_now = internal_clock_t::now();
           typename Clock::time_point  c_now = Clock::now();
-          wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
-          return Clock::now() < t ? cv_status::no_timeout : cv_status::timeout;
```
viboes: I believe this line is needed.

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -274,10 +306,9 @@ namespace boost
...
+          return wait_until(lock, s_now + ceil<nanoseconds>(t - c_now));
```
viboes: I believe that we should iterate until  Clock::now() >= t

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -286,99 +317,30 @@ namespace boost
...
-            wait_until(lock,
-                        nano_sys_tmpt(ceil<nanoseconds>(t.time_since_epoch())));
-            return steady_clock::now() < t ? cv_status::no_timeout :
-                                             cv_status::timeout;
-        }
-
-        template <class lock_type, class Clock, class Duration>
-        cv_status
+        template <class lock_type, class Duration, class Predicate>
+        bool
```
viboes: Why bool and not cv_status?

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -286,99 +317,30 @@ namespace boost
...
-            return steady_clock::now() - c_now < d ? cv_status::no_timeout :
-                                                   cv_status::timeout;
-        }
-
-        template <class lock_type>
-        inline cv_status wait_until(
-            lock_type& lock,
-            chrono::time_point<chrono::steady_clock, chrono::nanoseconds> tp)
+                lock_type& lock,
+                const chrono::time_point<internal_clock_t, Duration>& t,
```
viboes: Agreed

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -388,9 +350,9 @@ namespace boost
                 const chrono::duration<Rep, Period>& d,
                 Predicate pred)
         {
-          return wait_until(lock, chrono::steady_clock::now() + d, boost::move(pred));
+          return wait_until(lock, internal_clock_t::now() + d, boost::move(pred));
```
viboes: Agreed

include/boost/thread/pthread/condition_variable.hpp
brainwave64: Oops! Good catch. This is a copy/paste error.

brainwave64: Actually, no, this isn't a copy-paste error. In the existing code the functions that take a predicate return a bool instead of cv_status. I stayed consistent with that behavior.

viboes:
I've created the branch and I'm working on it:

You are right that we will need an absolute_internal_timespec.

I have defined: 
timespec_duration
real_timespec_timepoint
real_timespec_clock
mono_timespec_timepoint
mono_timespec_clock

and 

internal_timespec_timepoint
internal_timespec_clock

This makes much simpler things.

However we cannot convert between clocks until we  know the epoch, and  we don't know the epoch for steady_clock.

I've also extracted internal_clock_t to thread/detail/internal_clock.hpp  and renamed  detail::monotonic_pthread_cond_init to pthread::cond_init and extracted to pthread/pthread_helpers.hpp.

I've used those in thread, condition_variable, mutex and recursive_mutex.

I'm running the regression tests and will be able to have a first version tomorrow morning.

Things I must do:
* Inhibit pthread functions that don't have monotonic timed interface when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is defined
* Add a loop on functions doing something until a generic Clock::timepoint (some of them are buggy)
* Check each one of your proposed changes, in case I missed a bug in Boost.Thread code. 
* Simplify the code (later on)

Thanks for raising those issues.

include/boost/thread/detail/thread.hpp
viboes: Oh, I see. I suspect that we need to iterate until no_timeout or Clock::now() >= tp, isn't it?

brainwave64: The reason I took out and avoided code that loops until Clock::now() >= tp is because, if Clock::now() is the system clock and the system clock jumps, we could end up timing out at the wrong time.

My thought is that, even if someone passes us a system_clock time, if we have access to steady_clock (because BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled), we should convert from system_clock to steady_clock and then use steady_clock under the covers.

The following doesn't provide a 100% reliable conversion from a system_clock time to a steady_clock time, but I think it's better than not converting to steady_clock and then sleeping or timing out at the wrong time due to a system clock jump. This is the conversion that I use throughout the code.

steady_clock_time = ( system_clock_time - system_clock::now() ) + steady_clock::now()

include/boost/thread/detail/thread.hpp
viboes: RIght :)

include/boost/thread/detail/thread.hpp
viboes: RIght :)

include/boost/thread/pthread/condition_variable.hpp
brainwave64: See my comments above.

include/boost/thread/pthread/condition_variable.hpp
brainwave64: See my comments in detail/thread.hpp.

include/boost/thread/detail/thread.hpp
viboes: Ok, I see for timespec_to_internal_clock. The name is not a good one. 

I don't think the user wants to wait a delay, otherwise it would had used a duration.
He want to join until a specific system_time. If the time is changed while joining, either the time would already be reached and then we need to return or the time is not reached yet, and then we need to iterate. 

The current implementation is not good neither.

brainwave64:
That all sounds good.

The only issue I can think of is that you may break a lot of user's code if you inhibit pthread functions that don't take a monotonic timed interface when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled. I think it would be more useful if we kept those interfaces intact and just converted from system_clock to steady_clock. This doesn't break user's code but still gives them the benefit of a monotonic clock.

The following conversion function doesn't provide a 100% reliable conversion from a system_clock time to a steady_clock time, but I think it is better than either inhibiting system_clock functions or not using the monotonic clock even when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled.

steady_clock_time = ( system_clock_time - system_clock::now() ) + steady_clock::now()

Just something to consider...

include/boost/thread/detail/thread.hpp
```diff
@@ -157,7 +157,7 @@ namespace boost
     }
 namespace thread_detail {
 #ifdef BOOST_THREAD_USES_CHRONO
-#if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
+#if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC) || defined(BOOST_THREAD_PLATFORM_WIN32)
```
viboes: Why do you think that WIN32 has a monotonic clock?

brainwave64: https://github.com/boostorg/chrono/blob/develop/include/boost/chrono/config.hpp#L68

viboes:
> The only issue I can think of is that you may break a lot of user's code if you inhibit pthread functions that don't take a monotonic timed interface when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled. I think it would be more useful if we kept those interfaces intact and just converted from system_clock to steady_clock. This doesn't break user's code but still gives them the benefit of a monotonic clock.

I meant to inhibit the use of pthreads functions 

```
    #   if defined(BOOST_HAS_PTHREAD_DELAY_NP)
    #     if defined(__IBMCPP__) ||  defined(_AIX)
                  BOOST_VERIFY(!pthread_delay_np(const_cast<timespec*>(&ts.get())));
    #     else
                  BOOST_VERIFY(!pthread_delay_np(&ts.get()));
    #     endif
    #   elif defined(BOOST_HAS_NANOSLEEP)
                  nanosleep(&ts.get(), 0);
    #   else
```

> The following conversion function doesn't provide a 100% reliable conversion from a system_clock time to a steady_clock time, but I think it is better than either inhibiting system_clock functions or not using the monotonic clock even when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled.

I don't see from where you deduce that I want to "inhibiting system_clock functions or not using the monotonic clock even when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is enabled"

Sorry, the formula

steady_clock_time = ( system_clock_time - system_clock::now() ) + steady_clock::now()

is good, but I believe that we need to return cv_status::timeout only when `system_clock_time `is `>=`the new `system_clock::now()'.

include/boost/thread/detail/thread.hpp
viboes: Yes, it has a steady clock, but the WIN32 thread API Boost.Thread is using is a `system_clock` IIUC. I don't know Windows enough. Maybe there is an API that waits on a monotonic clock. 

So on windows, the `internal_clock_t` should be `system_clock` if I'm not wrong.

viboes: The same for MACOS. It has a steady_clock, but don't provide thread functions with MONOTONIC clocks, AFAIK.

Please, let me know if I'm missing something important.

include/boost/thread/detail/thread.hpp
viboes: I agree with the conversion. But I don't agree with the result of the function. The result should depend on whether the system timepoint has been reached or not.

include/boost/thread/pthread/mutex.hpp
viboes: I believe it is enough to check in addition  is defined or not BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC or define a new specific BOOST_THREAD_MACRO.

src/pthread/thread.cpp
```diff
@@ -438,21 +438,22 @@ namespace boost
                 if (boost::detail::timespec_ge(ts, boost::detail::timespec_zero()))
                 {
 
-    #   if defined(BOOST_HAS_PTHREAD_DELAY_NP)
+    #   if defined(BOOST_HAS_PTHREAD_DELAY_NP) && !defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
```
viboes: Agreed

src/pthread/thread.cpp
```diff
@@ -438,21 +438,22 @@ namespace boost
...
     #     else
                   BOOST_VERIFY(!pthread_delay_np(&ts));
     #     endif
-    #   elif defined(BOOST_HAS_NANOSLEEP)
+    #   elif defined(BOOST_HAS_NANOSLEEP) && !defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
                   //  nanosleep takes a timespec that is an offset, not
                   //  an absolute time.
                   nanosleep(&ts, 0);
     #   else
+                  const timespec ts2 = boost::detail::timespec_plus_internal_clock(ts);
```
viboes: Agreed

include/boost/thread/pthread/thread_data.hpp
```diff
@@ -292,13 +292,13 @@ namespace boost
 #endif
         inline void sleep(system_time const& abs_time)
         {
-          return boost::this_thread::hidden::sleep_until_realtime(boost::detail::to_timespec(abs_time));
+          boost::this_thread::hidden::sleep_until_realtime(boost::detail::to_timespec(abs_time));
```
viboes: Agreed

include/boost/thread/pthread/thread_data.hpp
```diff
@@ -292,13 +292,13 @@ namespace boost
...
         }
 
         template<typename TimeDuration>
         inline BOOST_SYMBOL_VISIBLE void sleep(TimeDuration const& rel_time)
         {
-            this_thread::sleep(get_system_time()+rel_time);
+          boost::this_thread::hidden::sleep_for(boost::detail::to_timespec(rel_time));
```
viboes: agreed

include/boost/thread/pthread/recursive_mutex.hpp
```diff
@@ -246,7 +272,7 @@ namespace boost
         template<typename TimeDuration>
         bool timed_lock(TimeDuration const & relative_time)
         {
-            return timed_lock(get_system_time()+relative_time);
+            return do_try_lock_until(detail::timespec_plus_internal_clock(relative_time));
```
viboes: agreed. We need to use internal_clock here

include/boost/thread/pthread/recursive_mutex.hpp
```diff
@@ -344,44 +370,34 @@ namespace boost
         }
     public:
 
-#endif
+#endif // !defined BOOST_USE_PTHREAD_RECURSIVE_TIMEDLOCK
 
 #if defined BOOST_THREAD_USES_DATETIME
         bool timed_lock(system_time const & abs_time)
         {
-            struct timespec const ts=detail::to_timespec(abs_time);
```
viboes: I will use here
```
real_timespec_timepoint const ts = abs_time;
```

include/boost/thread/pthread/recursive_mutex.hpp
```diff
@@ -344,44 +370,34 @@ namespace boost
...
-            return do_try_lock_until(ts);
+            return do_try_lock_until(detail::timespec_to_internal_clock(abs_time));
         }
 #endif
 #ifdef BOOST_THREAD_USES_CHRONO
         template <class Rep, class Period>
         bool try_lock_for(const chrono::duration<Rep, Period>& rel_time)
         {
-          return try_lock_until(chrono::steady_clock::now() + rel_time);
+          return try_lock_until(internal_clock_t::now() + rel_time);
```
viboes: agreed

include/boost/thread/pthread/recursive_mutex.hpp
```diff
@@ -344,44 +370,34 @@ namespace boost
...
         {
           using namespace chrono;
-          system_clock::time_point     s_now = system_clock::now();
+          internal_clock_t::time_point s_now = internal_clock_t::now();
           typename Clock::time_point  c_now = Clock::now();
           return try_lock_until(s_now + ceil<nanoseconds>(t - c_now));
         }
         template <class Duration>
-        bool try_lock_until(const chrono::time_point<chrono::system_clock, Duration>& t)
+        bool try_lock_until(const chrono::time_point<internal_clock_t, Duration>& t)
```
viboes: agreed

include/boost/thread/pthread/recursive_mutex.hpp
```diff
@@ -344,44 +370,34 @@ namespace boost
...
         {
-          using namespace chrono;
-          typedef time_point<system_clock, nanoseconds> nano_sys_tmpt;
```
viboes: I will change here

          typedef time_point<thread_detail::internal_clock_t, nanoseconds> nano_sys_tmpt;

viboes:
Hi again.

Here it is what I have done since yesterday https://github.com/boostorg/thread/tree/feature/timespec_clocks

Tested on MAC-OS only.

I'll continue replacing all the occurrences of `get_system_time()`.

I've surely forgotten some additional cases.

Once I've completed `get_system_time()` replacement, I would be interested in a test on linux with monotonic interface.

Of course any test on Windows is already welcome.

include/boost/thread/detail/thread.hpp
```diff
@@ -539,17 +529,14 @@ namespace thread_detail {
...
+        template <class Duration>
+        bool try_join_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& tp)
         {
-          using namespace chrono;
-          nanoseconds d = tp.time_since_epoch();
-          timespec ts = boost::detail::to_timespec(d);
-          return do_try_join_until(ts);
+          return do_try_join_until(boost::detail::to_timespec(tp.time_since_epoch()));
         }
 #endif
```
viboes: Here it is the endif of defined(BOOST_THREAD_PLATFORM_WIN32). So the next use of timespec is not good.
This code is common.

brainwave64: Yes, agreed. This change should be reverted, as it will break on Windows. My bad.

brainwave64:
Ah, I see. I misunderstood what you meant when you said "inhibit pthreads functions". Yes, I agree with doing that, and I think I already did that myself in v2/thread.hpp.

brainwave64:
I'd be happy to test your changes on Linux with a monotonic clock once you're ready for me to. Linux is my primary development platform. I'm not nearly as familiar with Windows either. :-(

brainwave64:
> I believe that we need to return cv_status::timeout only when system_clock_time is >=the new `system_clock::now()'.

I still think we should return cv_status::timeout only when do_wait_until() times out. Here's the scenario I'm trying to avoid:

- BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is defined, so do_wait_until() is using the steady clock.
- Some other code calls wait_until() in a loop until wait_until() returns cv_status::timeout, and it passes it a system time.
- wait_until() converts the system time to a steady time and calls do_wait_until().
- do_wait_until() times out after waiting the correct amount of time. But, while it was waiting, the system clock jumped.
- If wait_until() compares against the system clock to determine whether or not to return cv_status::timeout, it may incorrectly report that it did _not_ time out, even though do_wait_until() did time out, which will cause the other code's while loop to incorrectly call wait_until() again.

Does that make sense?

include/boost/thread/detail/thread.hpp
brainwave64: I agree that MAC OS, even though it has a steady clock, doesn't support a steady clock with pthread_cond_init and so the internal clock for MAC OS needs to be system_clock.

For Windows, however, it appears that all *wait_until() functions are internally converted to a duration. So using steady_clock on Windows as the internal clock should work just fine. And actually, because it converts everything to a duration, I don't think any Windows code should be using our internal clock typedefs or implementations at all. All Windows code should just be converting straight from a timepoint to a duration using the timepoint's now() function.

viboes:
@brainwave64 
I know it is surely not enoug https://github.com/boostorg/thread/commit/a6d5b5e576149838f4b14908f4b0c1a41fd01636, but please,could you give a try to this branch.

viboes:
Concerning the `cv_status`, returning `cv_status::timeout` when the type has not been reached seems to me an error. This is why I believe we should iterate again, until either `cv_status::no_timeout` or `t <=now()`.

I've yet to fix in my branch.

include/boost/thread/detail/thread.hpp
viboes: You are completely right. I don't know why, I believed we where using the system time.

There are some system_clock uses in win32/basic_timed_mutex.hpp
```
            template <class Rep, class Period>
            bool try_lock_for(const chrono::duration<Rep, Period>& rel_time)
            {
              return try_lock_until(chrono::steady_clock::now() + rel_time);
            }
            template <class Clock, class Duration>
            bool try_lock_until(const chrono::time_point<Clock, Duration>& t)
            {
              using namespace chrono;
              system_clock::time_point     s_now = system_clock::now();
              typename Clock::time_point  c_now = Clock::now();
              return try_lock_until(s_now + ceil<system_clock::duration>(t - c_now));
            }
            template <class Duration>
            bool try_lock_until(const chrono::time_point<chrono::system_clock, Duration>& t)
            {
              using namespace chrono;
              typedef time_point<chrono::system_clock, chrono::system_clock::duration> sys_tmpt;
              return try_lock_until(sys_tmpt(chrono::ceil<chrono::system_clock::duration>(t.time_since_epoch())));
            }
            bool try_lock_until(const chrono::time_point<chrono::system_clock, chrono::system_clock::duration>& tp)
```

My bad, surely :(

However I see that it was already the case for the DAtTime interface

```
            bool timed_lock(::boost::system_time const& wait_until)
            {
                if(try_lock())
                {
                    return true;
                }
                long old_count=active_count;
                mark_waiting_and_try_lock(old_count);

                if(old_count&lock_flag_value)
                {
                    bool lock_acquired=false;
                    void* const sem=get_event();

                    do
                    {
                        if(win32::WaitForSingleObjectEx(sem,::boost::detail::get_milliseconds_until(wait_until),0)!=0)
                        {
                            BOOST_INTERLOCKED_DECREMENT(&active_count);
                            return false;
                        }
                        clear_waiting_and_try_lock(old_count);
                        lock_acquired=!(old_count&lock_flag_value);
                    }
                    while(!lock_acquired);
                }
                return true;
            }

            template<typename Duration>
            bool timed_lock(Duration const& timeout)
            {
                return timed_lock(get_system_time()+timeout);
            }
```

I'll try to see how to fix this.

brainwave64:
> Concerning the cv_status, returning cv_status::timeout when the type has not been reached seems to me an error. This is why I believe we should iterate again, until either cv_status::no_timeout or t <=now().

Let me see if I can clarify where we disagree. Suppose a user gets the current system time, adds one second to it, and then passes that new timepoint to `wait_until()`. While `wait_until()` is waiting, the system time jumps backwards one second. There are three possible ways `wait_until()` could behave in this situation.

1. Wait for two seconds and then return `cv_status::timeout`. This is the current behavior with no monotonic clock.
2. Wait for one second and then return `cv_status::timeout`. This is the behavior of my PR code when the monotonic clock is enabled.
3. Wait for one second and then return `cv_status::no_timeout`. This is the behavior when the monotonic clock is enabled and we compare `t` against `Clock::now()` to determine which `cv_status` to return.

I think `#2` is better than `#3` because I think that 99% of the time, when a user passes us a system time that's in the future, they are expecting to wait for a fixed duration of time from now, and for the `cv_status` to indicate whether or not that fixed duration of time was reached.

However, I will concede that `#3` can be considered technically correct because the user passed us a system timepoint, not a steady timepoint, and so we should assume that they did so knowing the system time might jump. So I could go either way. I just think that `#2` will result in less unexpected behavior for users when the monotonic clock is enabled.

brainwave64:
I've tested out your branch and it looks pretty good so far. I had to make a few corrections to the code in order to get it to compile. I've included those in the diff below.

[compile_fixes.diff.txt](https://github.com/boostorg/thread/files/1273282/compile_fixes.diff.txt)

I've also included the results of my testing. The only tests that failed were the ones for functions that currently implement the `#3` behavior instead of the `#2` behavior (see my previous comment), since my tests are designed assuming `#2` behavior. Other than that everything now behaves much better with the monotonic clock enabled than it did in v1.65.

[results.txt](https://github.com/boostorg/thread/files/1273281/results.txt)

viboes:
I believe we should have `#1`. If the user is "expecting to wait for a fixed duration of time from now", it should use a duration. I believe this is independent of whether we have monotonic platform interface. See  http://talesofcpp.fusionfenix.com/post-15/rant-on-the-templated-nature-of-stdchrono, where Agustin shows that even with non-monotonic interface we can manage with clock adjustments by polling.

`#2` doesn't respect the contract.
`#3`is not valid, when the result of the first wait is successful. This is why I believe that we should iterate until success or the time-point is reached. 

If the user wants to #2, it can use a duration instead of a system time_point.

brainwave64:
Ok, you've convinced me. We should be doing `#1`, even when the underlying implementation is using a monotonic clock. That article was very helpful. Thank you for pointing me to it.

I'll update my test code to expect `#1` behavior when passed a system clock timepoint.

`#1` is a bit challenging to implement well when the monotonic clock is enabled, specifically in the case where the system clock jumps forward. I believe Agustin suggested implementing a maximum threshold in the polling loop so that `wait_until()` returns as soon as possible when the system clock jumps forward. Are you considering adding a threshold as well?

viboes:
Thanks for the test and the patch. Sorry for the code that contained compile errors, I was unable to compile it.

I don't understand why you need `real_timespec_timepoint` in

```
diff --git a/include/boost/thread/pthread/thread_data.hpp b/include/boost/thread/pthread/thread_data.hpp
index 275a6634..f8f3137e 100644
--- a/include/boost/thread/pthread/thread_data.hpp
+++ b/include/boost/thread/pthread/thread_data.hpp
@@ -244,7 +244,7 @@ namespace boost
         namespace hidden
         {
           void BOOST_THREAD_DECL sleep_for(const detail::timespec_duration& ts);
-          void BOOST_THREAD_DECL sleep_until_realtime(const detail::internal_timespec_timepoint& ts);
+          void BOOST_THREAD_DECL sleep_until_realtime(const detail::real_timespec_timepoint& ts);
         }
 
 #ifdef BOOST_THREAD_USES_CHRONO
@@ -265,7 +265,7 @@ namespace boost
           namespace hidden
           {
             void BOOST_THREAD_DECL sleep_for(const detail::timespec_duration& ts);
-            void BOOST_THREAD_DECL sleep_until_realtime(const detail::internal_timespec_timepoint& ts);
+            void BOOST_THREAD_DECL sleep_until_realtime(const detail::real_timespec_timepoint& ts);
           }
 
     #ifdef BOOST_THREAD_USES_CHRONO
@@ -292,13 +292,13 @@ namespace boost
 #endif
         inline void sleep(system_time const& abs_time)
         {
-          boost::this_thread::hidden::sleep_until_realtime(detail::internal_timespec_timepoint(abs_time));
+          boost::this_thread::hidden::sleep_until_realtime(detail::real_timespec_timepoint(abs_time));
         }
 
         inline BOOST_SYMBOL_VISIBLE void sleep(TimeDuration const& rel_time)
         {
-          boost::this_thread::hidden::sleep_until_realtime(detail::internal_timespec_clock::now() + detail::timespec_duration(rel_time));
+          boost::this_thread::hidden::sleep_until_realtime(detail::real_timespec_clock::now() + detail::timespec_duration(rel_time));
         }
 #endif // BOOST_THREAD_USES_DATETIME
     } // this_thread
```

I understand that I need to rename `sleep_until_realtime` to `sleep_until`. This was on my todo list, but I forget to do it.

I'm wondering if 

```
#if defined BOOST_THREAD_USES_DATETIME
      inline mono_timespec_timepoint(boost::system_time const& abs_time);
#endif
#if defined BOOST_THREAD_USES_CHRONO
      inline mono_timespec_timepoint(chrono::time_point<chrono::system_clock, chrono::nanoseconds> const& abs_time);
#endif
```

and should define only implicit constructions  from steady clocks
```
      inline mono_timespec_timepoint(chrono::time_point<chrono::steady_clock, chrono::nanoseconds> const& abs_time);
```

brainwave64:
Changing sleep_until_realtime was necessary in order to get it to compile, because in src/pthread/thread.hpp sleep_until_realtime is defined to take real_timespec_timepoint. But I think I should have changed src/pthread/thread.hpp instead of thread_data.hpp to take internal_timespec_timepoint.

brainwave64:
This is just a thought. I don't know if it's a good idea or not.

What if condition_variable, condition_variable_any, mutex, and recursive_mutex each contained two pthread_cond_t variables internally instead of one? One would be initialized with the steady clock, and one would be initialized with the system clock. All pthread_cond_signal() and pthread_cond_broadcast() calls would be sent to both condition variables. If the user provides us with a system timepoint, we would wait on the system-clock-initialized condition variable. If the user provides us with a duration or with a steady timepoint, we would wait on the steady-clock-initialized condition variable.

This would introduce additional runtime overhead into the implementation, so it may not be worth it. But it would eliminate the need for us to poll on the condition variable and thus would provide the most reliable behavior when the user provides a system timepoint and the system clock jumps backwards or forwards. And it would eliminate the need to convert from system timepoints to mono_timespec_timepoint.

viboes:
There are 3 way to implement the time-related functions depending on the interface of the platform

system time-point - MacOs

    - steady implementation

        - We should iterate by polling with a time-point that is  e.g. 1 sec after real_now

    - system implementation

        - We could use the system time directly and we risk to wait more than needed or

        - We should iterate by polling with a time-point that is  e.g. 1 sec after real_now

steady time-point - Linux

    - system implementation

        - we should iterate until success or the time reached using the mono_now + t - real_now and we risk to wait more than needed  or

        - We should iterate by polling with a time-point that is  e.g. 1 sec after mono_now

    - steady

        - We could use the steady time directly as there is no risk


duration - Windows
    - system implementation

        - we should iterate until success or the time reached using the mono_now + t - real_now and we risk to wait more than needed  or

        - We should iterate by polling with a time-point that is  e.g. 1 sec after mono_now

    - steady

        - We could use the steady duration directly as there is no risk

viboes:
I don't think having two condition variables is a good thing, but let me thing a little bit more on it.

waiting on a system time-point that can be adjusted is something would need to pool if we don't want to wait too long time.

viboes:
I'll try to implement `#1` during the week.

viboes:
I've updated the branch with some of your corrections.


I've not added yet the construction of mono_timespec_timepoint from real_timespec_timepoint.

Please, let me know of the compiler errors.

brainwave64:
From my testing with a recent Linux kernel, when the system time jumps forwards while we're waiting on a system-clock initialized condition variable, the condition variable returns early. Likewise, when the system clock jumps backwards, it waits the additional amount of time before returning. In both cases, `#1` behavior occurs without a risk of waiting longer than needed.

brainwave64:
Ok, I'll pull in your changes and re-compile.

brainwave64:
sleep_until_realtime needs to be fixed in src/pthread/thread.cpp.

brainwave64:
do_try_join_until_noexcept also needs to be fixed in src/pthread/thread.cpp. Essentially just replace all instances of `real_` with `internal_` in that file and it will compile on Linux.

viboes:
Sorry, I missed this file on the commit :(

viboes:
Wondering if there is not an cycle / pool issue here

```
#if defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC && defined BOOST_CHRONO_HAS_CLOCK_STEADY
    template <class Rep, class Period>
    void sleep_for(const chrono::duration<Rep, Period>& d)
    {
      using namespace chrono;
      if (d > duration<Rep, Period>::zero())
      {
        steady_clock::time_point c_timeout = steady_clock::now() + ceil<nanoseconds>(d);
        sleep_until(c_timeout);
      }
    }

    template <class Duration>
    inline BOOST_SYMBOL_VISIBLE
    void sleep_until(const chrono::time_point<chrono::steady_clock, Duration>& t)
    {
      using namespace chrono;
      sleep_for(t - steady_clock::now());
    }
```

While this finish by working, it is pooling. IMO, the last one should be removed.

brainwave64:
I think both of these functions should be removed, and

`#elif defined BOOST_THREAD_SLEEP_FOR_IS_STEADY`

should be replaced by

`#if defined(BOOST_THREAD_SLEEP_FOR_IS_STEADY) && !defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)`

That's what I did in my PR.

brainwave64:
And thanks for committing that file. It compiles without issue now!

brainwave64:
I've updated my tests to expect `#1` behavior and re-run the tests against the latest code in your branch. Hopefully this will help you see which functions need attention.

I've also gotten my tests running on Windows under Visual Studio 2015. I've included those results as well. I only had to make a couple of changes, also included below, to get those tests to compile and run.

[compile_fixes.diff.txt](https://github.com/boostorg/thread/files/1275243/compile_fixes.diff.txt)
[linux_results.txt](https://github.com/boostorg/thread/files/1275245/linux_results.txt)
[windows_results.txt](https://github.com/boostorg/thread/files/1275246/windows_results.txt)

viboes:
Please, could you upload the test file with the changes?
How do you calculate the expected time?
How much time the clock is back or forward?

Thanks for all

brainwave64:
Here is the updated test file. I also included the Makefiles I'm using in case those are helpful to you. Just be aware some of the paths are hard-coded for my environment.

[test_boost_sleep.cpp.txt](https://github.com/boostorg/thread/files/1275962/test_boost_sleep.cpp.txt)
[Makefile.linux.txt](https://github.com/boostorg/thread/files/1275964/Makefile.linux.txt)
[Makefile.win32.txt](https://github.com/boostorg/thread/files/1275965/Makefile.win32.txt)

For these tests, I get the current system or steady clock time (depending on the test), add 500 milliseconds, and then pass the result to the function being tested. Then, 100 milliseconds after calling the function, I change the system clock either forward or backward by 300 milliseconds.

For functions using the steady clock, they should wait 500 milliseconds regardless of how the system clock changed. For functions using the system clock, they should wait only 200 milliseconds when the system clock jumps forward by 300 milliseconds, and they should wait 800 milliseconds when the system clock jumps backward by 300 milliseconds. These are the expected values I check for.

viboes:
Thanks for the clarifications.

Glad to see that you could check on Windows as well.

Could you check, setting the internal_clock_t to steady_clock on windows?
```
          typedef chrono::steady_clock internal_clock_t;
```

viboes:
I've committed all your fixes and started to iterate while waiting until condition variables.
Please, let me know what are the results. Are there some improvements? Regressions?

brainwave64:
Changing the internal clock on Windows didn't change any of the test results. I did have to make one change to get it to compile, though. In `include/boost/thread/detail/thread.hpp`:
`typename Clock::duration   d = ceil<nanoseconds>(t-Clock::now());`
had to be changed to:
`typename Clock::duration   d = ceil<Clock::duration>(t-Clock::now());`
because the system clock on Windows uses a duration that is in units of 100-nanoseconds.

On Linux, your changes caused two additional tests to succeed that didn't before. There were no regressions:

* Testing boost::condition_variable::wait_until() with a system clock time:
    * While system clock jumps back: Succeeded
* Testing boost::condition_variable_any::wait_until() with a system clock time:
    * While system clock jumps back: Succeeded

These two tests still fail when the system clock jumps forward (it still waits too long), but that's to be expected since your changes didn't address that.

viboes:
Great. I 'll take in account the windows fix. Maybe we need something more specific, as the internal duration is nanoseconds on posix and milliseconds on windows, if I'm not wrong.

Concerning the yet failing jump forward tests:
* How much long?
* How would you address the jump forward?

brainwave64:
When the system clock jumps forward by 300 milliseconds, the wait_until() function still waits for 500 milliseconds. It should only wait for 200 milliseconds. So it waits 300 milliseconds too long.

There are only two solutions I can think of right now, but I'll think about it some more and see if I can come up with anything better.
* Add a maximum threshold (e.g. 100 milliseconds) on how long we wait so that we poll Clock::now() more frequently.
* Add a second internal pthread_cond_t variable that's initialized to use the system clock instead of the steady clock, and wait on that condition variable instead.

viboes:
Ok. I see. So when the system time is changed and we are waiting on a monotonic condition, there is no timeout or something else. Jumping backward is not a problem as we need to wait even more time, than requested on the monotonic clock.

I understand now why you susggested a second pthread_condition_t variable.

Polling every 100 milliseconds could be quite expensive. 

A third possibility could be to use timers. Instead of waiting for some time on a condition variable we could just arm a timer and wait on the condition (without time). When the timer timeouts it will notify the condition variable. The major problem is that this needs additional memory for the timer. 

I believe that we could implement the polling first and see how it works.
Then we could implement the 2 pthread_cond_t variables. 

After all, maybe the solution with a system condition variable wouldn't be so bad, if it is ensured that the wait on a  pthread_cond_t will finish when the system time is changed. Do you have any pointer where this feature is documented (required)?

viboes:
Could you share the results, when  BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is not defined?

viboes:
More on the timer solution. It is of course much more complex than I believed originally. We need cancelation and know that the system time has been changed :( 

All this should be provided by pthread_cond_t on system clock if wait finish when the system time changes.

brainwave64:
I don't have any documentation that specifies how pthread_cond_t should behave when the system time jumps forward. All I have is my testing on Linux that indicates that it will return early like we want it to.

brainwave64:
This is an interesting read. It looks like Mike Crowe started down the road of updating glibc so that a single pthread_cond_t variable could be used with multiple clocks via a new pthread_cond_timedwaitonclock_np function. That's essentially the same method as having two pthread_cond_t variables, only more efficient.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861

brainwave64:
I updated my tests again to test the sleep functions both when they are called from within a boost thread and when they are called from outside of a boost thread. When they are called from outside of a boost thread, they use the no_interruption_point implementations in `src/pthread/thread.cpp`, which weren't being tested before.

Here are the latest results run against the current code in feature/timespec_clocks. As would be expected, enabling BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC has no effect on the Windows results.

[test_boost_sleep.cpp.txt](https://github.com/boostorg/thread/files/1278597/test_boost_sleep.cpp.txt)
[linux_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1278596/linux_results_no_monotonic.txt)
[linux_results_with_monotonic.txt](https://github.com/boostorg/thread/files/1278598/linux_results_with_monotonic.txt)
[windows_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1278599/windows_results_no_monotonic.txt)
[windows_results_with_monotonic.txt](https://github.com/boostorg/thread/files/1278600/windows_results_with_monotonic.txt)

viboes:
Thanks for reporting. Yes, windows is independent of `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`

These results are weird
```
boost::condition_variable::wait_for():
    While system clock remains stable: Succeeded
    While system clock jumps back:     FAILED: Expected a wait of 500 ms but actual wait was 801 ms (too high)
    While system clock jumps forward:  FAILED: Expected a wait of 500 ms but actual wait was 200 ms (too low)
boost::condition_variable::wait_until(), system clock time:
    While system clock remains stable: Succeeded
    While system clock jumps back:     Succeeded
    While system clock jumps forward:  Succeeded
```

`wait_for(d)` is defined as `wait_until(system_clock::now()+d)`.

If `wait_until` succeeds while jumping, why `wait_for` doesn't? Do you have an explanation?

brainwave64:
The test for wait_until() with a system clock time expects wait_until() to wait 800 milliseconds when the clock jump backwards and 200 milliseconds when the clock jumps forward. When BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is disabled, that's exactly what it does, so the test succeeds.

But wait_for() is expected to wait 500 milliseconds regardless of whether or not the system clock jumps, which is does not. Thus those tests fail.

viboes:
Thanks for the bugzilla  link. Very instructive.

BTW, it would be interesting to know if the wait until/for returned no_timeout/timeout.
IIUC, it is correct to return timeout only if the time was reached or the duration elapsed.
So for the tests that fail with (too low) the result should be no_timeout, and for the test that fail with (too high) or succeed the result should be timeout.

Any failure at the interface level will be an error, while waiting more or less will be a QOI. I'm not saying that I don't want to improve the QOI (the probe is we are working on this), but before addressing the QOI, I would like to be sure we have a conforming implementation.

Please, could you add this information on the report? 

Thanks in advance for all your help.

viboes:
Ok, understood for the difference between wait_until and wait_for.

brainwave64:
Another compile error I just found in `include/boost/thread/v2/thread.hpp`
Change:
`ceil<thread_detail::internal_clock_t::::duration>(d)`
To:
`ceil<thread_detail::internal_clock_t::duration>(d)`

viboes:
Concerning 
```
 typename Clock::duration d = ceil<typename Clock::duration>(t-Clock::now());
```

I believe that it should be 
```
 typename Clock::duration d = t-Clock::now();
```

as the ceil is done already on wait_until, isn't it?

brainwave64:
Here are the updated test results you requested. I also added tests for the no_interruption_point versions of the sleep functions.

Currently only two tests on Windows return `no_timeout` when they fail with "too short". All of the other tests that fail with "too short" return `timeout`.

[test_boost_sleep.cpp.txt](https://github.com/boostorg/thread/files/1278983/test_boost_sleep.cpp.txt)
[linux_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1278984/linux_results_no_monotonic.txt)
[linux_results_with_monotonic.txt](https://github.com/boostorg/thread/files/1278985/linux_results_with_monotonic.txt)
[windows_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1278986/windows_results_no_monotonic.txt)

brainwave64:
> Concerning
> 
>  typename Clock::duration d = ceil<typename Clock::duration>(t-Clock::now());
> I believe that it should be
> 
>  typename Clock::duration d = t-Clock::now();
> as the ceil is done already on wait_until, isn't it?

Yes, you are correct.

viboes:
I have committed changes with the compile fixes you have reported and some that should fix the condition variable functions with chrono interface for linux-monotonic and linux-no-monotonic configurations.

brainwave64:
Here are the test results with your changes. The results don't indicate any regressions, and more of the tests pass now. I am going to update my test to not use times that are a multiple of 100 milliseconds, since that's the value you now use in your polling loop. I think we may be getting some false Passes as a result, but we'll see.

[linux_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1279130/linux_results_no_monotonic.txt)
[linux_results_with_monotonic.txt](https://github.com/boostorg/thread/files/1279131/linux_results_with_monotonic.txt)

brainwave64:
Changing the wait and jump times in my tests didn't change the results, so I think we're on the right track with the polling loop code. I looked at what you did and it looks really good to me. Two pthread_cond_t variables would provide somewhat better QOI, but with the cost of additional overhead, and it would still require polling for user-defined clocks (should anyone ever do that). So I think the polling solution with a 100 millisecond max is a good compromise.

The 100 milliseconds could be made configurable by the user if we're worried about performance.

One small thing to perhaps account for. We should only use chrono::steady_clock when BOOST_CHRONO_HAS_CLOCK_STEADY is defined. As such, we should not allow BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC to be defined if BOOST_THREAD_USES_CHRONO is defined but BOOST_CHRONO_HAS_CLOCK_STEADY is not defined.

viboes:
Yes, the 100 milliseconds should be configurable at compile time.

I was wondering if the polling period couldn't be an additional parameter of the time related functions, defaulted to the configured one, or have an additional overload. In this way the user is always able master the granularity of the possible polling.

viboes:
Concerning  BOOST_CHRONO_HAS_CLOCK_STEADY, yes, you are right, but currently Boost.Chrono defines always BOOST_CHRONO_HAS_CLOCK_STEADY If I'm not wrong.

viboes:
Committed "try a fix for sleep_for/sleep_until on linux."

brainwave64:
> Yes, the 100 milliseconds should be configurable at compile time.
> 
> I was wondering if the polling period couldn't be an additional parameter of the time related functions, defaulted to the configured one, or have an additional overload. In this way the user is always able master the granularity of the possible polling.

That's a good idea. I like it.

brainwave64:
> Concerning BOOST_CHRONO_HAS_CLOCK_STEADY, yes, you are right, but currently Boost.Chrono defines always BOOST_CHRONO_HAS_CLOCK_STEADY If I'm not wrong.

https://github.com/boostorg/chrono/blob/develop/include/boost/chrono/config.hpp#L85

The only possibility of BOOST_CHRONO_HAS_CLOCK_STEADY not being defined is if Boost is being compiled on a POSIX OS that doesn't support CLOCK_MONOTONIC. But I don't know if any such OSes exist anymore, so maybe it's safe to assume that CLOCK_MONOTONIC will always be present.

viboes:
Hrr, this means that we need to protect any use of steady_clock. :( 
And change the algorithm when not present.
The code on develop/release branches make use of steady_clock without protection and no one has reported any issue. 
I believe that I'll not protect it for the time being.

austin-beer:
> The code on develop/release branches make use of steady_clock without protection and no one has reported any issue.
> I believe that I'll not protect it for the time being.

Sounds good to me.

viboes:
I missed you comment about timeout-no_timeout. Thank for the change and report.
And glad to see that almost all the functions are reporting timeout :)
I'll address the Windows issues after the linux ones.

austin-beer:
Unfortunately, your sleep changes didn't fix any of the test results because of the presence of BOOST_THREAD_SLEEP_FOR_IS_STEADY and the fact that the `sleep*()` functions in `src/pthread/thread.cpp` get called when that is enabled. However, once I removed the BOOST_THREAD_SLEEP_FOR_IS_STEADY code, your changes caused additional tests to pass. See the file below for the changes I made and the subsequent test results.

I did have to fix an issue in the while loops. I added `d = t - Clock::now();` to them.

I also added a custom clock to my tests that is based on the system clock but with a different epoch. I just want to make sure that we don't break other clocks with these changes.

[fixes.diff.txt](https://github.com/boostorg/thread/files/1282015/fixes.diff.txt)
[test_boost_sleep.cpp.txt](https://github.com/boostorg/thread/files/1282014/test_boost_sleep.cpp.txt)
[linux_results_no_monotonic.txt](https://github.com/boostorg/thread/files/1282016/linux_results_no_monotonic.txt)
[linux_results_with_monotonic.txt](https://github.com/boostorg/thread/files/1282017/linux_results_with_monotonic.txt)

viboes:
Sorry for the missing `d = t - Clock::now();`. 
Maybe there are some additional issues, but I want to preserve `BOOST_THREAD_SLEEP_FOR_IS_STEADY` for platforms that have `BOOST_THREAD_SLEEP_FOR_IS_STEADY`  but `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`.

I recognize that there are some code to rework on thread_data. I'd need to check it.

viboes:
The custom clocks is an excellent idea.

viboes:
After inspection. I have not see major anything to change in thread_data.hpp.

Please redo a check, and let me know if some of your changes are still needed.

I've committed changes on mutex and  recursive_mutex.

viboes:
Sorry, I idn't do the changes in v2/thred.hpp yet

austin-beer:
I had to make just a couple of minor updates to your latest commit to get it to compile and run. The reason I changed `#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY` to `#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY && !defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` is because, without it, when `BOOST_THREAD_SLEEP_FOR_IS_STEADY` was defined, there ended up being two `sleep_until(const chrono::time_point<chrono::steady_clock, Duration>& t)` functions because internal_clock_t is a typedef of chrono::steady_clock.

[compile_fixes.diff.txt](https://github.com/boostorg/thread/files/1283375/compile_fixes.diff.txt)
[linux_results_no_monotonic_1.txt](https://github.com/boostorg/thread/files/1283377/linux_results_no_monotonic_1.txt)
[linux_results_with_monotonic_1.txt](https://github.com/boostorg/thread/files/1283376/linux_results_with_monotonic_1.txt)

Unfortunately there were still some sleep tests that failed. I took the liberty of making some additional changes that kept the BOOST_THREAD_SLEEP_FOR_IS_STEADY functionality (which should be kept, as you correctly pointed out), but that simplified the rest of the sleep code and caused the rest of the sleep tests to pass.

[sleep_fixes.diff.txt](https://github.com/boostorg/thread/files/1283387/sleep_fixes.diff.txt)
[linux_results_no_monotonic_2.txt](https://github.com/boostorg/thread/files/1283386/linux_results_no_monotonic_2.txt)
[linux_results_with_monotonic_2.txt](https://github.com/boostorg/thread/files/1283388/linux_results_with_monotonic_2.txt)

The mutex and recursive_mutex changes look good.

viboes:
The following lines in thread_data.hpp

```
 #ifdef BOOST_THREAD_USES_CHRONO
-        template <class Rep, class Period>
-        void sleep_for(const chrono::duration<Rep, Period>& d);
```

are there to prevent sleep_for to be instantiated for other duration than nanoseconds when this file is included, but v2/thread.hpp is not. Maybe I would need to add a comment.

austin-beer:
Ah, that makes sense. Yes, a comment would be helpful, since that reason wasn't obvious to me, and probably wouldn't be to others either. Thanks.

viboes:
I've not had time to analyze all your changes has needed.
I've committed the ones that were evident.

austin-beer:
No problem. Just let me know once you've had a chance to. Thanks!

viboes:
Hi, I've re-started with this issue.

Your patch has an issue, I believe.

sleep_for/sleep_until are interruption points and so we must use a condition variable.

I added no_interruption_point::sleep_for/sleep_until in order to provide sleep that are not interruption points.

What do you think?

viboes:
I have committed more changes.

https://github.com/boostorg/thread/commit/82eed2ca6d3b1573461ebcae19a9096eeff50799
https://github.com/boostorg/thread/commit/56016e834896aa09df44bff6ff6d166a2d6e1f47
https://github.com/boostorg/thread/commit/b6b6a6944a9526f49e70fe6d415e213c5b5f8824

Please, could we move this discussion to https://github.com/boostorg/thread/pull/142

Would you mind to report again all the results on this branch including windows?

I would appreciate if you create a sub-branch of https://github.com/boostorg/thread/pull/142 with the changes you want to propose. The best is to have a PR for each issue you find.

I've added Travis and AppVeyor CI so that we could see if the proposed changes break something.

I'm wondering that now that we are running the test on virtual machines, we could run your test in those machines. If you don't know how to update the Jamfile, I could do it.

viboes:
I'm closing this PR. See https://github.com/boostorg/thread/pull/142

PR #147, opened 2017-09-16
Change _pRawDllMain and related types to use HINSTANCE instead of HANDLE

Lastique:
In MSVC-9 MFC sources, _pRawDllMain and ExtRawDllMain accept HINSTANCE as
the first argument, not HANDLE. In strict mode these are different types,
which creates the potential for ODR errors. This commit resolves that
inconsistency.

Resolves https://svn.boost.org/trac10/ticket/12323.


viboes:
Andrey, thank you very much for your help.

PR #149, opened 2017-09-17
Fix/msvc12.0 devector issues

viboes:


pdimov:
It would be better to use `BOOST_MSVC` instead of `_MSC_VER` (as the latter is set by other compilers as well.)

In addition, `#if !BOOST_WORKAROUND(BOOST_MSVC, <= 1900)` might be a better check. (1900 is msvc-14.0, Visual Studio 2015.)

viboes:
Thanks Peter. I will fix it.

Issue #139, opened 2017-08-31
Clang on Windows build failures due to duplicated Windows API definitions


SSE4:
boost thread doesn't compile with Clang 4.0.1 on Windows, there are a lot of errors about duplicated Windows API definitions
```
In file included from libs\coroutine\src\windows\stack_traits.cpp:23:
In file included from .\boost/thread.hpp:13:
In file included from .\boost/thread/thread.hpp:12:
In file included from .\boost/thread/thread_only.hpp:15:
In file included from .\boost/thread/win32/thread_data.hpp:11:
.\boost/thread/win32/thread_primitives.hpp:173:55: error: conflicting types for 'CreateMutexA'
                __declspec(dllimport) void* __stdcall CreateMutexA(_SECURITY_ATTRIBUTES*,int,char const*);
                                                      ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um\synchapi.h:489:1: note: previous declaration is here
CreateMutexA(
^
In file included from libs\coroutine\src\windows\stack_traits.cpp:23:
In file included from .\boost/thread.hpp:13:
In file included from .\boost/thread/thread.hpp:12:
In file included from .\boost/thread/thread_only.hpp:15:
In file included from .\boost/thread/win32/thread_data.hpp:11:
.\boost/thread/win32/thread_primitives.hpp:174:55: error: conflicting types for 'CreateSemaphoreA'
                __declspec(dllimport) void* __stdcall CreateSemaphoreA(_SECURITY_ATTRIBUTES*,long,long,char const*);
                                                      ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um\winbase.h:3009:1: note: previous declaration is here
CreateSemaphoreA(
^
In file included from libs\coroutine\src\windows\stack_traits.cpp:23:
In file included from .\boost/thread.hpp:13:
In file included from .\boost/thread/thread.hpp:12:
In file included from .\boost/thread/thread_only.hpp:15:
In file included from .\boost/thread/win32/thread_data.hpp:11:
.\boost/thread/win32/thread_primitives.hpp:175:55: error: conflicting types for 'CreateEventA'
                __declspec(dllimport) void* __stdcall CreateEventA(_SECURITY_ATTRIBUTES*,int,int,char const*);
                                                      ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um\synchapi.h:530:1: note: previous declaration is here
CreateEventA(
^
[...]

Issue #99, opened 2016-09-06
queue_back_view: documentation mismatch


cinghiale:
The documentation says that `queue_back_view` supports  the [pull methods](https://github.com/boostorg/thread/blob/develop/doc/sync_queues_ref.qbk#L773) while they are not present in the implementation.

viboes:
Thanks for caching this error in the documentation.

A queue_back/front_view are useful to split the push and the pull functionalities.

I will try to fix it for the next version.

viboes:
https://github.com/boostorg/thread/commit/866ff746ae6b5c90c29dde7323fd5d6f36adbb30

PR #151, opened 2017-09-19
Updates to fix interruption-point inconsistencies.

austin-beer:
* Fixed the interruption-point versions of sleep_for/sleep_until() to always use condition variables.
* Fixed the no-interruption-point versions of sleep_for/sleep_until() to use pthread_delay_np or nanosleep whenever possible.
* Updated hidden::sleep_for() to always use a condition variable.
* Updated no_interruption_point::hidden::sleep_for() to use pthread_delay_np or nanosleep whenever possible.

NOTE: These changes also fix the two test failures with sleep_for/sleep_until when BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is disabled.

### DIFF
--- a/include/boost/thread/v2/thread.hpp
+++ b/include/boost/thread/v2/thread.hpp
@@ -23,33 +23,6 @@ namespace boost
     {
 #ifdef BOOST_THREAD_USES_CHRONO
 
-#if !defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
-    template <class Duration>
-    void sleep_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& t)
-    {
-      using namespace chrono;
-      mutex mut;
-      condition_variable cv;
-      unique_lock<mutex> lk(mut);
-      while (thread_detail::internal_clock_t::now() < t)
-      {
-        cv.wait_until(lk, t);
-      }
-    }
-    template <class Clock, class Duration>
-    void sleep_until(const chrono::time_point<Clock, Duration>& t)
-    {
-      using namespace chrono;
-      Duration d = t - Clock::now();
-      while (d > Duration::zero())
-      {
-        Duration d100 = (std::min)(d, Duration(milliseconds(100)));
-        sleep_until(thread_detail::internal_clock_t::now() + ceil<nanoseconds>(d100));
-        d = t - Clock::now();
-      }
-    }
-#endif
-
 // Use pthread_delay_np or nanosleep whenever possible here in the no_interruption_point
 // namespace because they do not provide an interruption point.
 #if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
@@ -96,6 +69,7 @@ namespace boost
     }
 
 #else
+
     template <class Rep, class Period>
     void sleep_for(const chrono::duration<Rep, Period>& d)
     {
@@ -103,6 +77,32 @@ namespace boost
       if (d > duration<Rep, Period>::zero())
       {
         sleep_until(steady_clock::now() + d);
+      }
+    }
+
+    template <class Duration>
+    void sleep_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& t)
+    {
+      using namespace chrono;
+      mutex mut;
+      condition_variable cv;
+      unique_lock<mutex> lk(mut);
+      while (thread_detail::internal_clock_t::now() < t)
+      {
+        cv.wait_until(lk, t);
+      }
+    }
+
+    template <class Clock, class Duration>
+    void sleep_until(const chrono::time_point<Clock, Duration>& t)
+    {
+      using namespace chrono;
+      Duration d = t - Clock::now();
+      while (d > Duration::zero())
+      {
+        Duration d100 = (std::min)(d, Duration(milliseconds(100)));
+        sleep_until(thread_detail::internal_clock_t::now() + ceil<nanoseconds>(d100));
+        d = t - Clock::now();
       }
     }
 
@@ -124,6 +124,7 @@ namespace boost
         cv.wait_until(lk, t);
       }
     }
+
     template <class Clock, class Duration>
     void sleep_until(const chrono::time_point<Clock, Duration>& t)
     {
@@ -136,6 +137,7 @@ namespace boost
         d = t - Clock::now();
       }
     }
+
     template <class Rep, class Period>
     void sleep_for(const chrono::duration<Rep, Period>& d)
     {
--- a/include/boost/thread/v2/thread.hpp
+++ b/include/boost/thread/v2/thread.hpp
@@ -23,6 +23,7 @@ namespace boost
     {
 #ifdef BOOST_THREAD_USES_CHRONO
 
+#if !defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
     template <class Duration>
     void sleep_until(const chrono::time_point<thread_detail::internal_clock_t, Duration>& t)
     {
@@ -47,7 +48,11 @@ namespace boost
         d = t - Clock::now();
       }
     }
-#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY && !defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif
+
+// Use pthread_delay_np or nanosleep whenever possible here in the no_interruption_point
+// namespace because they do not provide an interruption point.
+#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
 
     template <class Rep, class Period>
     void sleep_for(const chrono::duration<Rep, Period>& d)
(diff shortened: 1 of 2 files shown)


viboes:
Thanks for working on this.
Please, could you share the results fro the 3 variants linu-mono linux-no-mono and windows?

include/boost/thread/v2/thread.hpp
```diff
@@ -23,6 +23,7 @@ namespace boost
     {
 #ifdef BOOST_THREAD_USES_CHRONO
 
+#if !defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
```
viboes: IIUC, we have now just two variants:
* BOOST_THREAD_SLEEP_FOR_IS_STEADY
* NOT BOOST_THREAD_SLEEP_FOR_IS_STEADY

Could you rework the code to make this more evident.
See other comments how this can be possible

include/boost/thread/v2/thread.hpp
```diff
@@ -47,7 +48,11 @@ namespace boost
         d = t - Clock::now();
       }
     }
```
viboes: Add here the sleep_for. #99 #139

include/boost/thread/v2/thread.hpp
```diff
@@ -47,7 +48,11 @@ namespace boost
...
-#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY && !defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif
+
+// Use pthread_delay_np or nanosleep whenever possible here in the no_interruption_point
+// namespace because they do not provide an interruption point.
+#if defined BOOST_THREAD_SLEEP_FOR_IS_STEADY
```
viboes: Use #else instead

include/boost/thread/v2/thread.hpp
austin-beer: Yes, I can definitely do that. I was just trying to minimize code changes for the PR.

src/pthread/thread.cpp
```diff
@@ -437,25 +437,23 @@ namespace boost
...
                   BOOST_VERIFY(!pthread_delay_np(const_cast<timespec*>(&ts.get())));
     #     else
                   BOOST_VERIFY(!pthread_delay_np(&ts.get()));
     #     endif
-    #   elif defined(BOOST_HAS_NANOSLEEP) && !defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
+    #   elif defined(BOOST_HAS_NANOSLEEP)
                   nanosleep(&ts.get(), 0);
-    #   elif defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
+    #   else
+                  // Fall back to using a condition variable even though it does provide an interruption point.
```
viboes: IIUC, this function is called only in the BOOST_THREAD_SLEEP_FOR_IS_STEADY case and so the #else never happens, isn't it?

austin-beer: Yes, I believe that is correct. The #else should never get executed.

austin-beer:
Here are the test results from this PR.

[linux_results_austin_no_monotonic.txt](https://github.com/boostorg/thread/files/1315574/linux_results_austin_no_monotonic.txt)
[linux_results_austin_yes_monotonic.txt](https://github.com/boostorg/thread/files/1315575/linux_results_austin_yes_monotonic.txt)
[windows_results_austin_no_monotonic.txt](https://github.com/boostorg/thread/files/1315576/windows_results_austin_no_monotonic.txt)

The following two failures in the no-monotonic test are unavoidable. This is because hidden::sleep_for() has to use a condition variable (so that it has an interruption point), but it doesn't have access to a monotonic clock to see how much time has passed (unlike the sleep_for() that takes a Chrono time).

```
boost::this_thread::sleep(), posix duration:
    While system clock remains stable:        Expected: 580 ms, Actual: 580 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 580 ms, Actual: 810 ms, Returned: timeout, FAILED: TOO LONG
    While system clock jumps forward (short): Expected: 580 ms, Actual: 350 ms, Returned: timeout, FAILED: TOO SHORT
    While system clock jumps forward (long):  Expected: 580 ms, Actual: 110 ms, Returned: timeout, FAILED: TOO SHORT

 boost::this_thread::sleep(), no thread, posix duration:
    While system clock remains stable:        Expected: 580 ms, Actual: 580 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 580 ms, Actual: 810 ms, Returned: timeout, FAILED: TOO LONG
    While system clock jumps forward (short): Expected: 580 ms, Actual: 350 ms, Returned: timeout, FAILED: TOO SHORT
    While system clock jumps forward (long):  Expected: 580 ms, Actual: 110 ms, Returned: timeout, FAILED: TOO SHORT
```

viboes:
Thanks.

Shouldn't boost::this_thread::sleep(), posix duration: waits for 100ms? this will reduce the gap, isn't it?

austin-beer:
I'm afraid I don't quite understand your question. Could you rephrase the question?

viboes:
The current definition is

```
        inline BOOST_SYMBOL_VISIBLE void sleep(TimeDuration const& rel_time)
        {
          boost::this_thread::hidden::sleep_for(detail::timespec_duration(rel_time));
        }
```

Instead of using `boost::this_thread::hidden::sleep_for`, why not pool on a condition as we do in
https://github.com/boostorg/thread/pull/151/commits/f1dec7935fc2198702a2a5acf5512838628b384f#diff-52550ac33fe611a20ce1f27ce707309cR141

viboes:
I'll take this PR already, as aside the posix-time all in linux is ok.

BTW, 

I've see a regression on windows.

Before
```
boost::this_thread::sleep_until(), system time:
    While system clock remains stable:        Expected: 580 ms, Actual: 592 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 810 ms, Actual: 826 ms, Returned: timeout, Passed
    While system clock jumps forward (short): Expected: 350 ms, Actual: 373 ms, Returned: timeout, Passed
    While system clock jumps forward (long):  Expected: 110 ms, Actual: 217 ms, Returned: timeout, Passed
boost::this_thread::sleep_until(), custom time:
    While system clock remains stable:        Expected: 580 ms, Actual: 592 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 810 ms, Actual: 841 ms, Returned: timeout, Passed
    While system clock jumps forward (short): Expected: 350 ms, Actual: 357 ms, Returned: timeout, Passed
    While system clock jumps forward (long):  Expected: 110 ms, Actual: 233 ms, Returned: timeout, FAILED: TOO LONG
```

After
```
boost::this_thread::sleep_until(), system time:
    While system clock remains stable:        Expected: 580 ms, Actual: 607 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 810 ms, Actual: 841 ms, Returned: timeout, Passed
    While system clock jumps forward (short): Expected: 350 ms, Actual: 358 ms, Returned: timeout, Passed
    While system clock jumps forward (long):  Expected: 110 ms, Actual: 248 ms, Returned: timeout, FAILED: TOO LONG
boost::this_thread::sleep_until(), custom time:
    While system clock remains stable:        Expected: 580 ms, Actual: 592 ms, Returned: timeout, Passed
    While system clock jumps back:            Expected: 810 ms, Actual: 810 ms, Returned: timeout, Passed
    While system clock jumps forward (short): Expected: 350 ms, Actual: 358 ms, Returned: timeout, Passed
    While system clock jumps forward (long):  Expected: 110 ms, Actual: 124 ms, Returned: timeout, Passed
```

austin-beer:
I'm not sure if that is a true regression on Windows or not. I'm running all of my windows tests in a VM, and so I've seen up to 50 milliseconds of variability in the test results, which sometimes causes false results. I do have a dedicated Windows machine that I can use, but I don't have easy access to it at the moment. Once we start focusing on the Windows side of things, I plan to move to running my tests on that system.

austin-beer:
> The current definition is
> 
>         template<typename TimeDuration>
>         inline BOOST_SYMBOL_VISIBLE void sleep(TimeDuration const& rel_time)
>         {
>           boost::this_thread::hidden::sleep_for(detail::timespec_duration(rel_time));
>         }
> Instead of using boost::this_thread::hidden::sleep_for, why not pool on a condition as we do in
> f1dec79#diff-52550ac33fe611a20ce1f27ce707309cR141

We're able to poll every 100 milliseconds in other places in the code because we have a steady clock to compare against. In this function, all we have is the internal clock. When the internal clock is the system clock, there's no way to know when the system clock jumps, and so there's no way to wait the correct amount of time.

viboes:
Right, we are able to implement `sleep_for(chrono::duration)` because we have a steady::clock. posix-duration is another kind of duration, but we don't have a Date-Time steady clock.  
If we have BOOST_CHRONO we should be able to implement convert to chrono::duration, isn't it?
It will left the case when the user uses Boost.DateTime and not use Boost.Chrono.  I believe this is a really minor set of users.
What do you think?

viboes:
Don't worry for the Windows possible regression. You are surely right. I've observed much more  differences in the past on Windows.

austin-beer:
Yes, if we have BOOST_THREAD_USES_CHRONO we could convert to chrono::duration and that would fix the issue. And if BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC is defined that already fixes the issue. I'm not sure if it would be good to start mixing Chrono functions with DateTime functions. I like the fact that, currently, the two are completely separate from each other.

But most users do have BOOST_THREAD_USES_CHRONO and don't have BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC, so in that situation we should probably make sleep() work correctly by using chrono::duration under the covers.

viboes:
I agree that it would be desirable that Chrono and DateTime code stay independent, but from a pragmatic point of view, it would be a pity to don't provide the correct functionality when we have Boost.Chrono.

PR #153, opened 2017-09-19
Fixed sleep(TimeDuration const& rel_time)

austin-beer:
Fixed sleep(TimeDuration const& rel_time) when internal clock is not monotonic but CLOCK_MONOTONIC is available.

### DIFF
--- a/include/boost/thread/pthread/timespec.hpp
+++ b/include/boost/thread/pthread/timespec.hpp
@@ -203,7 +203,7 @@ namespace boost
       }
     };
 
-#if defined BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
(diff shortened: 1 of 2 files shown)


austin-beer:
Here are the test results for this PR. What do you think of my approach?

[linux_results_austin_no_monotonic.txt](https://github.com/boostorg/thread/files/1315856/linux_results_austin_no_monotonic.txt)
[linux_results_austin_yes_monotonic.txt](https://github.com/boostorg/thread/files/1315857/linux_results_austin_yes_monotonic.txt)

viboes:
Super. I've already forgotten that we have now timespec clocks :)

austin-beer:
Yep! I just remembered myself.

PR #154, opened 2017-09-19
Cleanup

austin-beer:
* Deleted a couple of unnecessary calls to internal_clock_t::now() in v2/thread.hpp.
* Deleted the hidden::sleep_until() functions, which are no longer being used.
* Deleted the condition_variable::do_wait_for() function, which is no longer being used.
* Deleted the sleep_mutex and sleep_condition variables in pthread/thread_data.hpp, which are no longer being used.

I re-ran the tests and, as expected, there are no changes to the results from this PR.


viboes:
Thanks again.

PR #155, opened 2017-09-20
Chrono-related optimizations

austin-beer:
* Fixed an issue where the results of chrono arithmetic operations weren't being assigned to a common type that's guaranteed to compile.
* Eliminated unnecessary conversions to chrono::nanoseconds.

As expected, none of the test results changed with these updates.

### DIFF
--- a/include/boost/thread/pthread/condition_variable.hpp
+++ b/include/boost/thread/pthread/condition_variable.hpp
@@ -297,14 +297,15 @@ namespace boost
                 const chrono::time_point<Clock, Duration>& t)
         {
           using namespace chrono;
-          Duration d = t - Clock::now();
-          if ( d <= Duration::zero() ) return cv_status::timeout;
-          d = (std::min)(d, Duration(milliseconds(100)));
+          typedef typename common_type<Duration, typename Clock::duration>::type CD;
+          CD d = t - Clock::now();
+          if ( d <= CD::zero() ) return cv_status::timeout;
+          d = (std::min)(d, CD(milliseconds(100)));
           while (cv_status::timeout == wait_until(lock, thread_detail::internal_clock_t::now() + d))
           {
               d = t - Clock::now();
-              if ( d <= Duration::zero() ) return cv_status::timeout;
-              d = (std::min)(d, Duration(milliseconds(100)));
+              if ( d <= CD::zero() ) return cv_status::timeout;
+              d = (std::min)(d, CD(milliseconds(100)));
           }
           return cv_status::no_timeout;
         }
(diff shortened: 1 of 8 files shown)


include/boost/thread/pthread/mutex.hpp
```diff
@@ -318,7 +318,7 @@ namespace boost
           Duration d = t - Clock::now();
           if ( d <= Duration::zero() ) return false;
           d = (std::min)(d, Duration(milliseconds(100)));
-          while ( ! try_lock_until(thread_detail::internal_clock_t::now() + ceil<nanoseconds>(d)))
+          while ( ! try_lock_until(thread_detail::internal_clock_t::now() + d))
```
viboes: Why not ceil here?

include/boost/thread/pthread/condition_variable_fwd.hpp
```diff
@@ -221,7 +220,7 @@ namespace boost
           Duration d = t - Clock::now();
           if ( d <= Duration::zero() ) return cv_status::timeout;
           d = (std::min)(d, Duration(milliseconds(100)));
-          while (cv_status::timeout == wait_until(lock, thread_detail::internal_clock_t::now() + ceil<nanoseconds>(d)))
+          while (cv_status::timeout == wait_until(lock, thread_detail::internal_clock_t::now() + d))
```
viboes: Why not ceil here?

include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -301,7 +300,7 @@ namespace boost
           Duration d = t - Clock::now();
           if ( d <= Duration::zero() ) return cv_status::timeout;
           d = (std::min)(d, Duration(milliseconds(100)));
-          while (cv_status::timeout == wait_until(lock, thread_detail::internal_clock_t::now() + ceil<nanoseconds>(d)))
+          while (cv_status::timeout == wait_until(lock, thread_detail::internal_clock_t::now() + d))
```
viboes: Why not ceil here?

include/boost/thread/pthread/mutex.hpp
austin-beer: The result of `thread_detail::internal_clock_t::now() + d` will be whatever type is common to both `internal_clock_t::duration` and `Duration`. Since `try_lock_until()` will accept any duration type, converting to nanoseconds is unnecessary. Does that make sense?

viboes: yes :)

PR #165, opened 2017-09-27
Added support for getting time on Windows to the timespec classes.

austin-beer:
On Windows, it seems to me that we need to eliminate the use of GitTickCount64() and the associated `timeout` class and replace it with our custom classes in timespec.hpp. If I'm duplicating work you've already done, then just ignore this PR. Otherwise, this PR allows us to use timespec.hpp on Windows.

### DIFF
--- a/include/boost/thread/detail/config.hpp
+++ b/include/boost/thread/detail/config.hpp
@@ -395,6 +395,17 @@
 #endif
 #endif
 
+#if defined(BOOST_THREAD_PLATFORM_WIN32)
+#define BOOST_THREAD_HAS_MONO_TIMESPEC
+#elif defined(BOOST_THREAD_MACOS)
+//#define BOOST_THREAD_HAS_MONO_TIMESPEC
+#else
+#include <time.h> // check for CLOCK_MONOTONIC
+#if defined(CLOCK_MONOTONIC)
+#define BOOST_THREAD_HAS_MONO_TIMESPEC
+#endif
+#endif
+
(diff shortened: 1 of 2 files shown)


MarcelRaad:
> On Windows, it seems to me that we need to eliminate the use of GitTickCount64() and the associated timeout class and replace it with our custom classes in timespec.hpp

Sorry I haven't followed this closely, but please make sure that there's still a way to use Boost.Thread without relying on `QueryPerformanceCounter`. It's more or less a random number generator on Windows Server 2003 with multiple CPUs (and on Windows XP with older chipsets, which I care less about). This is currently possible by using the Boost.DateTime interface and disabling Boost.Chrono usage (effectively reverting https://github.com/boostorg/thread/commit/d83d23bee0ad63eb6eeab6f51d2b49c871d367f9).

austin-beer:
Thanks for the info, Marcel. I wasn't aware of the issue with `QueryPerformanceCounter`, but that makes sense.

viboes:
Great. Thanks for this patch.

So the idea now is to move it to detail and use it at least for the generic  shared_mutex.hpp and for future.hpp and then  we could adapt the specific win32 files to use it, isn't it?

I'll check for the MacOs side on my side.

I'll commit it as soon as the CI is finished.

austin-beer:
Yep, that's exactly what I'm thinking.

I don't have access to macOS so I'm glad you do.

Thanks.

PR #166, opened 2017-09-28
Switching the Windows condition_variable and interruptible_wait functions over to using internal_timespec_clock.

austin-beer:
I ended up getting to this sooner than I'd planned.

Here are the results on Windows with this PR.

https://github.com/austin-beer/test-time-jumps/blob/master/windows_results_austin_no_generic_shared.txt
https://github.com/austin-beer/test-time-jumps/blob/master/windows_results_austin_yes_generic_shared.txt

### DIFF
--- a/include/boost/thread/detail/config.hpp
+++ b/include/boost/thread/detail/config.hpp
@@ -12,7 +12,6 @@
 #include <boost/detail/workaround.hpp>
 #include <boost/thread/detail/platform.hpp>
 
-//#define BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
 //#define BOOST_THREAD_DONT_PROVIDE_INTERRUPTIONS
 // ATTRIBUTE_MAY_ALIAS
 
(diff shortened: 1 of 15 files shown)


include/boost/thread/pthread/timespec.hpp
```diff
@@ -278,20 +303,52 @@ namespace boost
...
+        {
+          BOOST_ASSERT(0 && "Boost::Thread - QueryPerformanceCounter Internal Error");
+          return mono_timespec_timepoint(0);
+        }
+      }
+
+      long double ns = 1000000000.0L * pcount.QuadPart / freq.QuadPart;
+      return mono_timespec_timepoint(static_cast<boost::intmax_t>(ns));
+#else
+      // Use GetTickCount64() because it's more reliable on older
```
viboes: Does it means that Boost.Chrono should use this on those systems?

viboes:
Wow, this is a really big change and a big simplification of the code :)
I would need some time to digest all these change, in particular every thing concerning the windows changes.

I don't know why AppVeyor CI has not been launched on this PR.

viboes:
I found the AppVeyor link
https://ci.appveyor.com/project/viboes/thread/build/1.0.93-feature/timespec_clocks

Everything is Ok.

include/boost/thread/pthread/timespec.hpp
austin-beer: Yes, that is correct. Boost Chrono should probably provide an option so users can use GetTickCount64() instead of QueryPerformanceCounter() when using chrono::steady_time on older systems. Right now the only option those users have is to avoid using chrono::steady_time on those systems.

viboes:
The windows report is impressive. 
* No more no-timeout, 
* generic shared_mutex and 
* future 

are all ok :)

Only failing
* some timed_mutex
* some recursive_timed_mutex and
* some shared_mutex

I'll not have too much time before the C++ standard meeting mailing list deadline (16 Oct). I will really appreciate if you can work on this.

BTW, I'm looking for a co-maintainer for Boost.Thread ;-)

include/boost/thread/detail/config.hpp
```diff
@@ -386,6 +385,23 @@
...
 #if defined(BOOST_THREAD_PLATFORM_WIN32)
+  #define BOOST_THREAD_HAS_MONO_CLOCK
+  #define BOOST_THREAD_INTERNAL_CLOCK_IS_MONO
+#elif defined(BOOST_THREAD_MACOS)
+  //#define BOOST_THREAD_HAS_MONO_CLOCK
+#else
+  #include <time.h> // check for CLOCK_MONOTONIC
+  #if defined(CLOCK_MONOTONIC)
+    #define BOOST_THREAD_HAS_MONO_CLOCK
+    #if defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC)
```
viboes: We need a way to detect this. We need a program that succeeds if everything is there and works as expected (Something like https://github.com/boostorg/thread/blob/develop/build/has_atomic_flag_lockfree_test.cpp). 

Then we can use it as in

https://github.com/boostorg/thread/blob/develop/build/Jamfile.v2#L154

and

https://github.com/boostorg/thread/blob/develop/build/Jamfile.v2#L280

include/boost/thread/future.hpp
```diff
@@ -39,9 +39,7 @@
 #include <boost/thread/executor.hpp>
 #include <boost/thread/executors/generic_executor_ref.hpp>
 
-#if defined(BOOST_THREAD_PLATFORM_PTHREAD)
 #include <boost/thread/pthread/timespec.hpp>
```
viboes: We need to move this to detail

include/boost/thread/pthread/timespec.hpp
viboes: Issue added https://github.com/boostorg/chrono/issues/28

MarcelRaad: That would be great! To be exact, it's the GetTickCount64 emulation that works best on those systems. Unfortunately, GetTickCount64 itself has also only been introduced in Windows Vista.

PR #167, opened 2017-09-29
Fixes and Cleanup

austin-beer:
- Fixed build failures on Windows. The timespec struct is not supported by older versions of Visual Studio. I changed the internal representation inside of the *_timespec_timepoint classes to a boost::intmax_t representing the number of nanoseconds since the epoch.
- Fixed some functions that wouldn't execute at all if they were provided a negative time duration or an absolute time that was in the past. From what I understand, they should instead execute once and then return immediately.
- Moved pthread/timespec.hpp to detail/timespec.hpp.
- Deleted detail/internal_clock.hpp and moved the seven relevant lines into detail/timespec.hpp. This keeps all of the internal clock declarations in one place.
- Renamed thread_detail::internal_clock_t to detail::internal_chrono_clock to be consistent with and yet clearly differentiated from detail::internal_timespec_clock.
- Removed "using namespace chrono" to eliminate ambiguious namespace resolution when referencing detail::internal_chrono_clock.
- Re-enabled a few tests on Windows that had previously been disabled. I want to see whether or not they still need to be disabled.

I'm hoping this PR will fix the AppVeyor failures that the last PR introduced. I have been building with Visual Studio 2015 but I switched over to Visual Studio 2010 to better catch issues like this in the future.


include/boost/thread/pthread/condition_variable.hpp
```diff
@@ -112,17 +111,18 @@ namespace boost
 #endif
         int cond_res;
         {
+            timespec const& ts = timeout.getTs();
```
viboes: I find this weird. if getTs() return now a timespec, why not just define a variable as in
```
const timespec ts =  timeout.getTs();
```
?

I wonder if we shouldn't store timespec on pthread systems and intmax_t in windows in internal_timespec_duration. Would this change add a lot of of conditional compilation in detail/timespec .hpp?

In addition it has no sense to name it timespec.hpp if it can be something else.

What do you think?

viboes:
Ok, I will add an old version of msvc to check this kind backward compatibility issues.

include/boost/thread/pthread/condition_variable.hpp
viboes: What about platform_clock, platform_duration and platform_timepoint?

The file could be renamed to platform_time.hpp

viboes:
There are a lot of error in Windows and also on MacOs :(

austin-beer:
Yep, I'm working on those as we speak.

include/boost/thread/pthread/condition_variable.hpp
austin-beer: Agreed. It is weird. I'll change it.

I think you're right that we should probably store timespec on pthread and intmax_t on Windows. That would be the most efficient. I don't think it will add too much conditional compilation.

austin-beer: And I like the *platform* names instead of *timespec*.

include/boost/thread/pthread/mutex.hpp
```diff
@@ -266,8 +266,7 @@ namespace boost
         // fixme: Shouldn't this functions be located on a .cpp file?
         bool do_try_lock_until(detail::internal_timespec_timepoint const &timeout)
         {
-          const timespec ts = timeout.getTs();
-          int const res=pthread_mutex_timedlock(&m,&ts);
+          int const res=pthread_mutex_timedlock(&m,&timeout.getTs());
```
viboes: I find this weir. Passing the address of the returned value.

test/sync/mutual_exclusion/locks/unique_lock/cons/move_ctor_upgrade_lock_for_pass.cpp
```diff
@@ -30,6 +30,7 @@ boost::upgrade_mutex m;
 
 int main()
 {
+#if defined(BOOST_THREAD_PLATFORM_PTHREAD)
```
viboes: Now that win32 works with the pthread/shared_mutex generic implementation this should works. What was wrong?

include/boost/thread/pthread/mutex.hpp
austin-beer: That's what the original code did. getTs() passes a const reference so it's valid and saves an unnecessary copy of the variable.

test/sync/mutual_exclusion/locks/unique_lock/cons/move_ctor_upgrade_lock_for_pass.cpp
austin-beer: I'd have to go back and re-run it again to remember the exact error. It was something related to not being able to copy (or maybe move) between mutexes or locks of different types.

include/boost/thread/detail/timespec.hpp
```diff
@@ -137,15 +137,15 @@ namespace boost
       explicit real_timespec_timepoint(boost::intmax_t const& ns) : value(ns) {}
 
 #if defined BOOST_THREAD_USES_DATETIME
-      real_timespec_timepoint(boost::system_time const& abs_time)
+      explicit real_timespec_timepoint(boost::system_time const& abs_time)
```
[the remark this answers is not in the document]
austin-beer: I was getting an error awhile back related to Visual Studio 2010 not finding the correct constructor. But that may have been due to other issues that are now fixed. I can remove it again and see if it still works fine?

test/sync/mutual_exclusion/locks/unique_lock/cons/move_ctor_upgrade_lock_for_pass.cpp
austin-beer: Here's an example of the error that occurs in these tests with Visual Studio 2010. I'm not sure what the problem is. Do you have any ideas?

```
sync\mutual_exclusion\locks\unique_lock\cons\move_ctor_upgrade_lock_try_pass.cpp(34) : error C2664: 'boost::unique_lock<Mutex>::unique_lock<boost::try_to_lock_t>(Mutex &,const TimeDuration &)' : cannot convert parameter 1 from 'boost::upgrade_lock<Mutex>' to 'boost::upgrade_mutex &'
        with
        [
            Mutex=boost::upgrade_mutex,
            TimeDuration=boost::try_to_lock_t
        ]
        and
        [
            Mutex=boost::upgrade_mutex
        ]
```

austin-beer: Oh, I see what's going on, although I'm not sure why. In `lock_types.hpp`, Visual Studio 2010 is choosing this constructor:
```
    unique_lock(Mutex& m_,TimeDuration const& target_time);
```
instead of this constructor:
```
    unique_lock(Mutex& m_, try_to_lock_t);
```

austin-beer: Oh.....in `lock_types.hpp`, this is the constructor it needs, but it's not available on Windows:

```
#if defined(BOOST_THREAD_PLATFORM_PTHREAD)
    // Conversion from upgrade locking
    unique_lock(BOOST_THREAD_RV_REF_BEG upgrade_lock<mutex_type> BOOST_THREAD_RV_REF_END ul, try_to_lock_t)
    : m(0),is_locked(false)
    {
      if (BOOST_THREAD_RV(ul).owns_lock())
      {
        if (BOOST_THREAD_RV(ul).mutex()->try_unlock_upgrade_and_lock())
        {
          m = BOOST_THREAD_RV(ul).release();
          is_locked = true;
        }
      }
      else
      {
        m = BOOST_THREAD_RV(ul).release();
      }
    }
#endif
```

austin-beer: I got it...the #if defines were added by you and I just failed to remove them. That's why the tests are failing. I'll take care of it. Sorry for the spam.

include/boost/thread/pthread/mutex.hpp
viboes: It us just I'm not used to see this kind of construct. It is just me.

test/sync/mutual_exclusion/locks/unique_lock/cons/move_ctor_upgrade_lock_for_pass.cpp
viboes: No problem. I know that this is working now in develop, so I know we just did something wrong.

viboes:
The issue from 
```
libs/thread/test/sync/mutual_exclusion/sync_pq/pq_single_thread_pass.cpp(58): test 'diff < milliseconds(600) && diff > milliseconds(500)' failed in function 'void test_pull_for()'
```

is not your fault.

The implementation/interface of sync_priority_queue is wrong. I'll try to fix it.

Hrr, there are too much wrong time related interfaces in Boost.Thread.

viboes:
I don't remember why I dded such time-related functions as the C++ proposal it is based on (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0260r1.html) don't provides such functions.

viboes:
If you believe your PR is ready to merge I'll do.

viboes:
Could give a last try with
```
  BOOST_TEST(diff < milliseconds(650) && diff > milliseconds(500));
```

I suspect that now that we could be pooling (at least in MacOs) waiting for 500ms could be very easily wait for more than 600ms.

viboes:
As you see I've already merged your PR.

austin-beer:
Thanks. I've been out of the office all weekend so I wasn't able to take care of this. Thanks for taking a look at the pq_single_thread_pass issues.

PR #174, opened 2017-10-08
Try to fix Travis failures with the built-in g++ 4.8

pdimov:
This first commit just enables Travis testing with the default g++ compiler, I'll add another commit that will actually take care of the issue later. Don't merge yet. :-)


pdimov:
This seems to have done the trick.

viboes:
Thanks, this is what I wanted to suggest.

PR #178, opened 2017-10-12
Update tests to consistently pass on Windows (hopefully)

austin-beer:
Increased the time thresholds on a bunch of tests to be more forgiving so they pass consistently on Windows.

Reduced the time thresholds on a bunch of tests that were TOO forgiving.

### DIFF
--- a/test/sync/conditions/condition_variable/wait_for_pass.cpp
+++ b/test/sync/conditions/condition_variable/wait_for_pass.cpp
@@ -44,19 +44,17 @@ void f()
     test1 = 1;
     cv.notify_one();
     Clock::time_point t0 = Clock::now();
-    int count=0;
-    while (test2 == 0 && cv.wait_for(lk, milliseconds(250)) == boost::cv_status::no_timeout)
-      count++;
+    Clock::time_point t = t0 + milliseconds(250);
+    while (test2 == 0 && cv.wait_for(lk, t - Clock::now()) == boost::cv_status::no_timeout) {}
     Clock::time_point t1 = Clock::now();
     if (runs == 0)
     {
-      assert(t1 - t0 < milliseconds(250));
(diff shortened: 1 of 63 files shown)


viboes:
I was wondering if we couldn't have less constraining duration only for windows, but we can do it once we have a green CI.

austin-beer:
Yes, I think that would be a good idea. Do all the tests by chance share a common include file where we could set the threshold once and then use it in each of the tests?

austin-beer:
Is it possible to cancel this AppVeyor job (https://ci.appveyor.com/project/viboes/thread/build/1.0.147-feature/timespec_clocks) since it's been superseded by this one (https://ci.appveyor.com/project/viboes/thread/build/1.0.148-feature/timespec_clocks)?

viboes:
I've stopped the job.

I wonder if we shoudln't print the difference. I have observed that sometimes we have less time than expected, no more. This can be due to time slicing; there is some time between the call to now and the call to wait.

austin-beer:
All of the changes I've made allow the tests to take less time than expected without reporting errors. It's only when the tests take more than 200 milliseconds longer than expected that they report an error.

Unfortunately I've seen that sometimes 200 milliseconds isn't long enough. I think the issue is that AppVeyor is using heavily-loaded VMs and so the tests sometimes run even slower than they do on a normal Windows system.

austin-beer:
The Travis CI failed due to a timing issue unrelated to these changes. My guess is it will pass if run again.

The AppVeyor Visual Studio failures were due to issues unrelated to these changes. It seems like AppVeyor isn't always the most reliable CI platform.

The AppVeyor MinGW failures are due to legitimate issues that we need to address, but that should probably be done in a separate PR, since they are unrelated to this PR.

I'll add code to each test in this PR to set the timeout threshold based on whether we're running on Windows.

PR #179, opened 2017-10-13
Fix shared_mutex and future time jump tests

austin-beer:
This PR is composed of two commits.

The first commit simply reverts `pthread/shared_mutex.hpp` and `future.hpp` back to the versions in `develop`, which reverts all of the changes we've made to these two files to date.

The second commit re-fixes the time jump issues by using the predicate versions of the wait functions, which is a simpler solution than the one we were using before and which works even with the changes we made to avoid losing notifications.

### DIFF
--- a/include/boost/thread/future.hpp
+++ b/include/boost/thread/future.hpp
@@ -38,8 +38,6 @@
 #include <boost/thread/executor.hpp>
 #include <boost/thread/executors/generic_executor_ref.hpp>
 
-#include <boost/thread/detail/platform_time.hpp>
-
 #if defined BOOST_THREAD_FUTURE_USES_OPTIONAL
 #include <boost/optional.hpp>
 #else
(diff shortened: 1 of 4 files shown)


austin-beer:
Here are the test results from this PR.

https://github.com/austin-beer/test-time-jumps/blob/master/linux_results_austin_no_monotonic.txt
https://github.com/austin-beer/test-time-jumps/blob/master/linux_results_austin_yes_monotonic.txt
https://github.com/austin-beer/test-time-jumps/blob/master/windows_results_austin_no_generic_shared.txt
https://github.com/austin-beer/test-time-jumps/blob/master/windows_results_austin_yes_generic_shared.txt

The only fixable failures that still need to be fixed after this PR are in:
```
win32/basic_timed_mutex.hpp
win32/basic_recursive_mutex.hpp
win32/shared_mutex.hpp
```

viboes:
Thanks a lot for almost fixing everything.
There is some hope we could have this for boost-1.66.

PR #180, opened 2017-10-13
Add and fix sleep functions

austin-beer:
- Added no_interruption_point::sleep() functions to pthreads to be
  consistent with Windows.
- Fixed an issue where the no_interruption_point::sleep_*() functions
  were still interruptible on Windows.

In order to support these changes, I moved all of the Chrono sleep functions that were in `v2/thread.hpp` into `pthread/thread_data.hpp` and `win32/thread_data.hpp`.


viboes:
Good idea.

PR #188, opened 2017-10-26
Cleanup

austin-beer:
These are a series of commits that been sitting in my local workspace for a couple of weeks now. I was going to wait until after I'd finished the win32 changes before submitting them, but I decided to go ahead and submit them now to get them out of the way.

See the commit log for the list of changes I made.


viboes:
Great we start to see green run in windows :)

austin-beer:
A few green runs, at least. :)

austin-beer:
FYI, the 64-bit Windows builds are failing due the unit tests exceeding the 260 character filepath limit on Windows. Every filepath includes `\msvc-14.1\debug\address-model-64\threadapi-win32\threading-multi`. Do you know if there is an easy way to shorten this? On the 32-bit builds the filepaths don't have `\address-model-64` which is just enough to keep them under the 260 character limit so they don't fail.

viboes:
Thanks for the info. `/threadapi-win32` was added in order to do cross compiling.

No I don't know how to shorten them. I'll create an issue with Boost.Build.

viboes:
Steven Watanabe said on Boost ML
"
--abbreviate-paths

http://www.boost.org/build/doc/html/bbv2/reference/buildprocess.html#bbv2.reference.buildprocess.targetpath
"

viboes:
I've added --abbreviate-paths

austin-beer:
--abbreviate-paths sounds like the solution we need. Thanks for adding that.

viboes:
Everything is green :)

Thanks for all your efforts

austin-beer:
Glad to help. Which AppVeyor build are you looking at that's all green?

PR #190, opened 2017-11-01
Sync queue time jump fixes

austin-beer:
Please review this PR closely when you get the chance. I made several changes/bug fixes while reviewing the code in addition to fixing the time jump issues. See the change log for details.

Here are the before and after time jump test results for this PR:
https://github.com/austin-beer/test-time-jumps/blob/master/linux_results_fixed1_default.txt
https://github.com/austin-beer/test-time-jumps/blob/master/linux_results_austin1_default.txt

### DIFF
--- a/include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
+++ b/include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
@@ -55,7 +55,7 @@ namespace detail
     inline void close();
 
     inline underlying_queue_type underlying_queue() {
-      lock_guard<mutex> lk(mtx_);
+      unique_lock<mutex> lk(mtx_);
       return boost::move(data_);
     }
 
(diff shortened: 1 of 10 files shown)


include/boost/thread/shared_mutex.hpp
```diff
@@ -13,13 +13,20 @@
 #include <boost/thread/detail/config.hpp>
 #if defined(BOOST_THREAD_PLATFORM_WIN32)
 #if defined(BOOST_THREAD_PROVIDES_GENERIC_SHARED_MUTEX_ON_WIN)
+#if defined(BOOST_THREAD_V2_SHARED_MUTEX)
+#include <boost/thread/v2/shared_mutex.hpp>
```
viboes: I don't remember why I've two generic implementation. Wondering what are the differences?

include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
```diff
@@ -55,7 +55,7 @@ namespace detail
     inline void close();
 
     inline underlying_queue_type underlying_queue() {
-      lock_guard<mutex> lk(mtx_);
```
viboes: While you are right that the cost is minimal, I don't see why to pay for if we can avoid it.

include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
```diff
@@ -11,6 +11,8 @@
 //
 //////////////////////////////////////////////////////////////////////////////
 
+#include <boost/bind.hpp>
```
viboes: Do you mind to create another PR with this change only?

austin-beer: I can do that. Which change specifically are you referring to?

include/boost/thread/shared_mutex.hpp
austin-beer: The v2/shared_mutex.hpp implementation was one you added back in 2013 based on an implementation by H. Hinnant. I don't know if it's a better implementation than the existing one or not.

include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
viboes: These ones https://github.com/boostorg/thread/pull/190/commits/5371b04154e73de9abbab1267b26676e9f748059

You can add also the shared_mutex changes https://github.com/boostorg/thread/pull/190/commits/c5ab2d6865e001c408de6b962a27d4f3f0304982, but we need to see what is the added value.

austin-beer:
Replacing this PR with PR #191.

PR #196, opened 2017-12-20
Fix final time jump issues

austin-beer:
The following implementations have been fixed:

- include/boost/thread/win32/basic_timed_mutex.hpp
- include/boost/thread/win32/basic_recursive_mutex.hpp
- include/boost/thread/win32/shared_mutex.hpp
- include/boost/thread/v2/shared_mutex.hpp

### DIFF
--- a/include/boost/thread/v2/shared_mutex.hpp
+++ b/include/boost/thread/v2/shared_mutex.hpp
@@ -153,7 +153,9 @@ public:
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/condition_variable.hpp>
 #include <boost/thread/mutex.hpp>
+#ifdef BOOST_THREAD_USES_CHRONO
 #include <boost/chrono.hpp>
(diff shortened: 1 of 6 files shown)


viboes:
even if there are multiple issues with macos, I've merged it as you are unable to fix them.

viboes:
I would like you to do a PR with the test you have. maybe a complete folder for these test would be the best.

austin-beer:
It looks like all of the MAC OS test failures are due to not timing out when we should. My guess is there's an issue with how we're calculating or using the time on MAC OS.

I'll work on a PR to merge my test code into a folder in the Boost.Thread library. I probably won't get to it until after the holiday break, though.

viboes:
I'll check my changes on the timespec for MacOs (but the test were ok after my change) and are ok in my MacOs :(

No problem. I will have a holiday break also.

BTW, thanks again.

PR #200, opened 2018-01-12
Add CPU count for VxWorks

kuhlenough:
OS specific API required

### DIFF
--- a/src/pthread/thread.cpp
+++ b/src/pthread/thread.cpp
@@ -27,6 +27,10 @@
 #include <unistd.h>
 #endif
 
+#if defined(__VXWORKS__) 
+#include <vxCpuLib.h>
+#endif 
+
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/lexical_cast.hpp>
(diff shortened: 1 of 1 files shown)


src/pthread/thread.cpp
```diff
@@ -556,6 +560,18 @@ namespace boost
...
+        cpuset_t set =  ::vxCpuEnabledGet();
+  #ifdef __DCC__
+        int i;
+        for( i = 0; set; ++i)
+        {
+           set &= set -1;
+        }
+        return(i);
+  #else  
+        return (__builtin_popcount(set) );
```
viboes: Does it means that when we use VxWorks, we have always a compiler that has this intrinsic?
With with compiler has you tested?

kuhlenough: VxWorks is packaged with one of 4 validated compilers depending on the board support package and processor, so I know that Intel ICC, GCC, and llvm/clang have __builtin_popcount() and  Diab compiler (__DCC_) does not.  I didn't find a Boost macro for popcount() but I agree its a cleaner solution.

viboes: Okay, I accept the PR. We will adapt if someone else need something else.

viboes: Thanks for the PR.

PR #202, opened 2018-01-22
future.hpp:

shinobu-x:
1) Removed redundant access specifiers and unused variables in future_deferred_shared_state
2) check lock before unlock

### DIFF
--- a/include/boost/thread/future.hpp
+++ b/include/boost/thread/future.hpp
@@ -1007,10 +1007,8 @@ namespace boost
         template<typename Rp, typename Fp>
         struct future_deferred_shared_state: shared_state<Rp>
         {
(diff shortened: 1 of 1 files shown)


include/boost/thread/future.hpp
```diff
@@ -1133,7 +1126,8 @@ namespace boost
                 {
                     for(count_type_portable i=0;i<count;++i)
                     {
-                        locks[i].unlock();
+                        if (locks[i].owns_lock())
```
viboes: Hi, have you found a case where the lock where not locked?

shinobu-x: @viboes I've completely missed this function is expected to `unlock`

PR #203, opened 2018-01-26
Not *_FUNTION_* but *_FUNCTION_*

shinobu-x:


viboes:
Thanks for catching this type

PR #204, opened 2018-01-28
Missing destructor

shinobu-x:


viboes:
Instead of adding one we could just remove all these trivial destructors, isn't it?

shinobu-x:
@viboes true :)

PR #205, opened 2018-01-29
fix mingw build

WSoptics:
The following change is necessary for me to cross-compile using `mingw`.


viboes:
Thanks for the PR.

Please, could you show what was wrong on mingw build? This will help me to understand why do we need this change.

WSoptics:
Sorry, we don't need it.  It was caused by a dangling `winapi/wait.hpp` from another boost version in our repository and my stupidity.  Sorry for having wasted your time!

viboes:
Don't worry

PR #206, opened 2018-02-03
CallableType should be private

shinobu-x:


viboes:
What is the added value? This is a detail implementation class.
I believe that if the member `f` is public, his type should be public.

shinobu-x:
Should this type also be public?

https://github.com/shinobu-x/thread/blob/e60eca5212177386c271ce5046def824e5794123/include/boost/thread/future.hpp#L3176-L3182

viboes:
I don't know where you have posted
>Should this type also be public?
>
>https://github.com/shinobu-x/thread/blob/e60eca5212177386c271ce5046def824e5794123/include/boost/thread/future.hpp#L3176-L3182

But the answer is yes.

PR #207, opened 2018-02-13
Final Cleanup

austin-beer:


include/boost/thread/detail/thread.hpp
```diff
@@ -718,7 +713,7 @@ namespace boost
     };
 
 #ifdef BOOST_THREAD_PLATFORM_PTHREAD
-    thread::id thread::get_id() const BOOST_NOEXCEPT
+    inline thread::id thread::get_id() const BOOST_NOEXCEPT
```
viboes: There is no need to add inline here. It is better to add it on the declarations AFAIK.

include/boost/thread/pthread/thread_data.hpp
```diff
@@ -264,7 +264,7 @@ namespace boost
         }
 
         template<typename TimeDuration>
-        void sleep(TimeDuration const& rel_time)
+        inline void sleep(TimeDuration const& rel_time)
```
viboes: No need for inline here, it is a template

include/boost/thread/detail/thread.hpp
austin-beer: According to this FAQ, it is better to place inline next to the definition instead of the declaration. Do you think this FAQ is correct, as far as you know? https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

viboes:
Thanks for working on this PR.
I don't agree with all the inline changes. I believe that it is better to declare the function inline than defining it inline. I know that most compilers accept it, but I believe this is not portable.

include/boost/thread/pthread/thread_data.hpp
austin-beer: I'm doing a little more research on the inline keyword as we speak. I want to make sure I understand it fully before backing out the inline changes.

austin-beer: But I agree. I want to see this branch merged back into develop as soon as possible.

include/boost/thread/detail/thread.hpp
viboes: I was not aware of this. I agree that the code is more readable.
As it is in the isocpp.org site, lets say that it should be portable. I've no time to do search.

include/boost/thread/pthread/thread_data.hpp
viboes: I propose to merge it. We could improve the inline later if needed.
Do you agree?

austin-beer: Yes, I agree. We can merge now and improve later if need be.

austin-beer: Just an FYI...

Unlike non-template functions, template functions that are defined in multiple translation units (i.e. in a header file) do *not* need to use the inline keyword to satisfy the One Definition Rule (which I'm sure you're well aware of). However, I haven't found any documentation online to indicate that these template functions are treated as inline from an *optimization* perspective. So I think that template function definitions that we want the compiler to inline to improve performance should still be explicitly marked as inline.

viboes: I believe the compiler is free to do whatever concerns with optimizations independently of whether we add inline or not. I believe it is as for the member functions defined directly on the declaration.

The fact that I don't see this as an idiom in any library (including) let me think that it is not useful. 
Have you see some libraries using it?
Have you see a case where adding inline has as consequence that the compiler inline the template and don't if not added?

A very different thing could be to use some kind of FORCE_INLINE.  But I would prefer to not go this way, as it is very dependent on the compiler.

austin-beer: Ok, you make a good point. No, I haven't seen this idiom anywhere else so I think you are correct that we shouldn't do it here either. I can remove it in a future PR after we merge back to develop.

PR #208, opened 2018-02-19
Should retrun rvalue

shinobu-x:


viboes:
Please, don't provide PR to master, but develop brach.

PR #210, opened 2018-02-20
Remove unnecessary inline keyword from templated functions

austin-beer:


Lastique:
What is the point of this change?

austin-beer:
It reverts some of the changes I made in 3b7ca04408f8e5b45ed5f128ae3bd4ee27d883e5. I added inline as a hint to the compiler's optimizer, but after discussing it with Vicente I realized that such hints are not necessary. Thus the inline keyword is only required by function definitions that appear in header files outside of class/struct definitions in order to ensure that the One Definition Rule is not violated.

Issue #209, opened 2018-02-20
Legal problem with `win32/thread_primitives.hpp`


pdimov:
A third-party legal review has identified a problem as reported here:

https://lists.boost.org/Archives/boost/2018/02/241435.php

regarding

https://github.com/boostorg/thread/blob/526c72cb4b8e3cb14d6f8e008006ef36b2be2c39/include/boost/thread/win32/thread_primitives.hpp#L89-L110

introduced by

https://github.com/boostorg/thread/commit/04c53415fd0adc1921d419c54a1e5ec4b19917ea

The easiest way to fix that is to rewrite `GetTickCount64emulation` in terms of `boost::atomic_uint64_t`. As pointed out in

https://lists.boost.org/Archives/boost/2018/02/241453.php

an implementation already exists in Boost.Log:

https://github.com/boostorg/log/blob/1cc577cbf5fae8f55c71c4493a7ef89027bd85dc/src/timestamp.cpp#L66-L86

viboes:
I missed the comment on this commit.

could someone provide a PR?

pdimov:
@Lastique perhaps?

Lastique:
I will make a PR shortly.

viboes:
Thank you very much.

PR #211, opened 2018-02-20
Rewrite GetTickCount64 emulation implementation.

Lastique:
This is to resolve the possible license violation as the previous
implementation has been taken from StackOverflow and was not licensed
under the Boost Software License. The new implementation was adopted from
Boost.Log:

https://github.com/boostorg/log/blob/1cc577cbf5fae8f55c71c4493a7ef89027bd85dc/src/timestamp.cpp#L66-L86

The legal issue has been raised in:

https://lists.boost.org/Archives/boost/2018/02/241453.php

Fixes https://github.com/boostorg/thread/issues/209.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -284,6 +284,7 @@ alias thread_sources
       win32/thread.cpp
       win32/tss_dll.cpp
       win32/tss_pe.cpp
+      win32/thread_primitives.cpp
       future.cpp
     : ## requirements ##
       <threadapi>win32
--- a/include/boost/thread/detail/platform_time.hpp
+++ b/include/boost/thread/detail/platform_time.hpp
@@ -429,7 +429,7 @@ inline FP init_steady_clock(kern_return_t & err)
 #else
       // Use GetTickCount64() because it's more reliable on older
       // systems like Windows XP and Windows Server 2003.
-      win32::ticks_type msec = win32::GetTickCount64_()();
+      win32::ticks_type msec = win32::gettickcount64();
       return mono_platform_timepoint(msec * 1000000);
 #endif
 #elif defined(BOOST_THREAD_CHRONO_MAC_API)
--- a/include/boost/thread/win32/thread_primitives.hpp
+++ b/include/boost/thread/win32/thread_primitives.hpp
@@ -18,10 +18,9 @@
 #include <boost/detail/interlocked.hpp>
 
 #include <boost/winapi/config.hpp>
+#include <boost/winapi/basic_types.hpp>
 #include <boost/winapi/semaphore.hpp>
-#include <boost/winapi/dll.hpp>
 #include <boost/winapi/system.hpp>
-#include <boost/winapi/time.hpp>
 #include <boost/winapi/event.hpp>
 #include <boost/winapi/thread.hpp>
 #include <boost/winapi/get_current_thread.hpp>
@@ -48,8 +47,7 @@ namespace boost
         {
             typedef ::boost::winapi::HANDLE_ handle;
             typedef ::boost::winapi::SYSTEM_INFO_ system_info;
-            typedef unsigned __int64 ticks_type;
-            typedef ::boost::winapi::FARPROC_ farproc_t;
+            typedef ::boost::winapi::ULONGLONG_ ticks_type;
             unsigned const infinite=::boost::winapi::INFINITE_;
             unsigned const timeout=::boost::winapi::WAIT_TIMEOUT_;
             handle const invalid_handle_value=::boost::winapi::INVALID_HANDLE_VALUE_;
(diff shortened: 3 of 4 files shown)


ned14:
Unfortunately this PR would be a regression as currently written. The original bug report that I fixed was timers failing on a computer which sleeps for several months at a time. This PR does not detect and directly handle multiple 43 day system timer wraps as the original does.

Lastique:
I've updated the implementation to mitigate the GetTickCount wraparound problem if the user doesn't call GetTickCount64 for extended periods of time.

ned14:
Eh, surely a sleeping computer wouldn't call your refresh routine? Thus you get multiple wraps and the timer fires late?

Lastique:
Does `GetTickCount` keep going while the machine is sleeping?

ned14:
Yes

Lastique:
Oh well. Then if the machine sleeps for too long (more than ~49 days) the algorithm will miss one or more wraparounds. The current implementation doesn't seem to protect against that either, only it returns a bogus 0xFFFFFFFF value. Out of the two I'd prefer mine result since it at least somehow correlates with reality.

Anyways, let Vicente decide what to do with this PR. I'm fine if it's rejected.

ned14:
I honestly can't remember whether we specially handle the 32 bit GetTickCount64 emulation wrapping and returning -1. I do remember that we had a very eager bug reporter who did a lot of testing and verified the bug was fixed. Can't remember if that was on 64 bit only, and we threw WinXP to the wind.

Actually, seeing as we've now dropped WinXP as a supported platform, we could just axe the emulation altogether? Vista and later implement GetLastCount64(), including on x86 platforms.

Lastique:
The special return value is 0xFFFFFFFF, not -1 (note that GetTickCount64 return ULONGLONG, the 64-bit unsigned integer). The only place that calls GetTickCount64 (https://github.com/boostorg/thread/blob/develop/include/boost/thread/detail/platform_time.hpp#L432) does not check for this value. And the genuine GetTickCount64 has no special return values to indicate errors, so checking for 0xFFFFFFFF would be a wrong thing to do anyway.

> Actually, seeing as we've now dropped WinXP as a supported platform, we could just axe the emulation altogether?

I'm not aware of Boost.Thread dropping support for XP. Boost in general certainly didn't drop XP.

If Boost.Thread does not support pre-Vista then I agree, calling GetTickCount64 directly is more preferable.

viboes:
Thanks to both for working on this. Sorry, I know nothing about Windows.

IIUC neither the current code nor the proposed patch are perfect solutions to the problem, isn't it?
Could we state clearly what are the advantages/limitations of each approach?
Niall, do you believe that Andrey solution is not satisfactory?

IIUC, we use the Windows ::GetTickCount64 if available, isn't it?
In the absence of 64 bits counters, do we really need an emulation? Couldn't we just have a restriction? It is too late?

ned14:
WinXP EOLed in 2014. WinVista EOLed last year, in 2017.

If you feel that continuing to support WinXP is excessive, both of us agree that the GetTickCount64 emulation function ought to be removed entirely.

So decide on that first, then we go from there.

Lastique:
> IIUC, we use the Windows ::GetTickCount64 if available, isn't it?

Yes, both versions do.

viboes:
Thanks to both.

We need to support any version we support today, and in particular WinXP.
Nial, my concern wasn't not about not supporting WinXp, but on not using an emulation for GetTickCount64, but this was surely a bad idea. 

I believe that the PR limitations are acceptable. 
Niall if you think the PR has any major issues, please tell us which.
If no one is completely against, I'll merge this PR 25/02/2018.

Vicente

Kojoley:
I am getting an error on Appveyor that I could not reproduce locally.
```
.\boost/thread/win32/thread_primitives.hpp(73): error C2143: syntax error: missing ')' before '*'
.\boost/thread/win32/thread_primitives.hpp(73): error C2143: syntax error: missing ';' before '*'
.\boost/thread/win32/thread_primitives.hpp(73): error C2059: syntax error: ')'
.\boost/thread/win32/thread_primitives.hpp(73): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
.\boost/thread/win32/thread_primitives.hpp(74): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
.\boost/thread/win32/thread_primitives.hpp(74): error C2365: 'boost::detail::win32::detail::gettickcount64_t': redefinition; previous definition was 'function'
.\boost/thread/win32/thread_primitives.hpp(73): note: see declaration of 'boost::detail::win32::detail::gettickcount64_t'
.\boost/thread/win32/thread_primitives.hpp(74): error C2146: syntax error: missing ';' before identifier 'gettickcount64'
```

Update: It looks like Boost.Thread CI is experiencing the same issue.

Lastique:
See https://github.com/boostorg/thread/pull/214.

Issue #225, opened 2018-07-30
warning: cast between incompatible function types from FARPROC_ to gettickcount64_t


kivadiu:
While building 1.68.0 beta 1 on x86_64 cross-compiling to Windows 32 bits with mingw-w64 + g++ 8.2.0, I get the following warning:

```
"i686-w64-mingw32-g++-8.2.0"   -I/softs/win32-mingw-8.2.0/release/iconv/include -I/softs/win32-mingw-8.2.0/release/gettext/include -I/softs/win32-mingw-8.2.0/release/bzip2/include -I/softs/win32-mingw-8.2.0/release/zlib/include -I/softs/win32-mingw-8.2.0/release/jpeg-turbo/include -I/softs/win32-mingw-8.2.0/release/xz/include -I/softs/win32-mingw-8.2.0/release/tiff/include -I/softs/win32-mingw-8.2.0/release/png/include -std=c++14 -O2 -DNDEBUG -m32 -mthreads -O3 -finline-functions -Wno-inline -Wall -pedantic -march=i686 -Wextra -Wno-long-long -Wno-unused-parameter -Wunused-function -pedantic -DBOOST_ALL_NO_LIB=1 -DBOOST_ASIO_NO_DEPRECATED -DBOOST_CHRONO_STATIC_LINK=1 -DBOOST_FILESYSTEM_NO_DEPRECATED -DBOOST_LOG_WITHOUT_EVENT_LOG -DBOOST_SYSTEM_NO_DEPRECATED -DBOOST_SYSTEM_STATIC_LINK=1 -DBOOST_THREAD_BUILD_LIB=1 -DBOOST_THREAD_USES_CHRONO -DBOOST_THREAD_WIN32 -DBOOST_USE_WINDOWS_H -DNDEBUG -DWIN32_LEAN_AND_MEAN  -I"." -c -o "bin.v2/libs/thread/build/gcc-8.2.0/release/link-static/target-os-windows/threadapi-win32/threading-multi/win32/thread_primitives.o" "libs/thread/src/win32/thread_primitives.cpp"
libs/thread/src/win32/thread_primitives.cpp: In function 'boost::detail::win32::ticks_type boost::detail::win32::{anonymous}::get_tick_count_init()':
[...]

PR #227, opened 2018-08-08
Fixed warning cast between incompatible function types from FARPROC_ to gettickcount64_t

kivadiu:
On Windows, GetProcAddress must be cast to something useful which
generates a warning with gcc. This patch silents the warning.

Fixes issue #225

### DIFF
--- a/src/win32/thread_primitives.cpp
+++ b/src/win32/thread_primitives.cpp
@@ -93,8 +93,24 @@ ticks_type WINAPI get_tick_count_init()
     boost::winapi::HMODULE_ hKernel32 = boost::winapi::GetModuleHandleW(L"kernel32.dll");
     if (hKernel32)
     {
+      // GetProcAddress returns a pointer to some function. It can return
+      // pointers to different functions, so it has to return something that is
+      // suitable to store any pointer to function. Microsoft chose FARPROC,
+      // which is int (WINAPI *)() on 32-bit Windows. The user is supposed to
+      // know the signature of the function he requests and perform a cast
+      // (which is a nop on this platform). The result is a pointer to function
+      // with the required signature, which is bitwise equal to what
+      // GetProcAddress returned.
+      // However, gcc >= 8 warns about that.
+#if defined(BOOST_GCC) && BOOST_GCC >= 80000
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
         boost::detail::win32::detail::gettickcount64_t p =
             (boost::detail::win32::detail::gettickcount64_t)boost::winapi::get_proc_address(hKernel32, "GetTickCount64");
+#if defined(BOOST_GCC) && BOOST_GCC >= 80000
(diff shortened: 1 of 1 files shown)


src/win32/thread_primitives.cpp
```diff
@@ -93,8 +94,24 @@ ticks_type WINAPI get_tick_count_init()
...
+      // pointers to different functions, so it has to return something that is
+      // suitable to store any pointer to function. Microsoft chose FARPROC,
+      // which is int (WINAPI *)() on 32-bit Windows. The user is supposed to
+      // know the signature of the function he requests and perform a cast
+      // (which is a nop on this platform). The result is a pointer to function
+      // with the required signature, which is bitwise equal to what
+      // GetProcAddress returned.
+      // However, gcc warns about that.
+#if BOOST_COMP_GNUC
+#pragma GCC diagnostic push
```
Lastique: Pragma diagnostic push/pop appeared only since gcc 4.6. You should check for the compiler version.

Also, you don't need Boost.Predef to detect gcc. You can use `BOOST_GCC` from Boost.Config. Also, it is quite possible that clang also emits a similar warning, you can check for it with `BOOST_CLANG`.

src/win32/thread_primitives.cpp
```diff
@@ -93,8 +94,24 @@ ticks_type WINAPI get_tick_count_init()
...
+#pragma GCC diagnostic ignored "-Wcast-function-type"
```
Lastique: This warning is only documented (and likely implemented) since gcc 8.

kivadiu: That's right, thanks. I will change that.

src/win32/thread_primitives.cpp
kivadiu: I cannot check for clang as I do not cross-compile to Windows with clang but I will use BOOST_GCC.

kivadiu:
I am very surprised by test failures in completely unrelated code.
For example, it fails with msvc-11 (but not with 9, 10 and 12) at [test/threads/thread/members/try_join_for_pass.cpp:147](https://ci.appveyor.com/project/viboes/thread/build/1.0.263-develop/job/tg32bf3yxrq4e5xl):
```
boost::thread t0( (G()));
BOOST_TEST(t0.joinable());
t0.try_join_for(boost::chrono::milliseconds(250));
try
{
  t0.join();
  BOOST_TEST(false); // line 147: test failure here
}
catch (boost::system::system_error& e)
{
  BOOST_TEST(e.code().value() == boost::system::errc::invalid_argument);
}
```

Is it possible that the test server is so heavily loaded that 250 ms are not enough and `t0.join()` does not fail as expected?

Lastique:
It is possible, as with any other time-based test.

viboes:
Andrey is right. This kind of tests can unfortunately fail.

kivadiu:
How do you normally decide if a patch is harmless then?

viboes:
I need to check the kind of issue :(

Issue #233, opened 2018-09-14
Compiling for ARM32/64 on VS2017 broken


DjArt:
Due assign in interlocked_read.hpp:
```
        inline void* interlocked_read_acquire(void* volatile* x) BOOST_NOEXCEPT
        {
            void* const res=
#if defined(_M_ARM64)
                __iso_volatile_load64((const volatile __int64*)x);
#else
                __iso_volatile_load32((const volatile __int32*)x);
#endif
            BOOST_THREAD_DETAIL_COMPILER_BARRIER();
            __dmb(0xB); // _ARM_BARRIER_ISH, see armintr.h from MSVC 11 and later
            BOOST_THREAD_DETAIL_COMPILER_BARRIER();
            return res;
        }
```

viboes:
Please, could you add the report you are getting?

DjArt:
@viboes, sorry for the long wait. Error C2440: int can't be converted to void*.

viboes:
Humm, if the function doesn't returns a void * why do we want to store it on a voir *?

DjArt:
`void*` used as platform-unspecified(by size) pointer. So, I think that's better variant for saving original idea of code.

viboes:
Oh, I see. 

So IIUC, __iso_volatile_load64 returned always a __int64 and __iso_volatile_load32 returned always __int32. 

And this size corresponds to the size of a pointer depending on whether _M_ARM64 is defined or not.

How it is that this worked before, Andrei, what do you think?

DjArt:
I think, or this code was untested before, or MSVC for ARM has become more strict then before. x86/x86_64 compiler doesn't consider it as mistake.

viboes:
I believe the code needs a static assert about the size of both types. Could you add it, please?

PR #234, opened 2018-09-14
Fixing compiling on VS2017 for ARM & ARM64

DjArt:
https://github.com/boostorg/thread/issues/233


viboes:
I've finished by accepting the PR.

The result of the function is used always by testing if it is != 0. So the cast should not make the things worst. I'd appreciate a PR for the static assert on the equality of the size of the types.

PR #238, opened 2018-10-03
Simplify thread specific storage

### LINKED ISSUE #236: UBSan false positive in thread_specific_ptr
`./b2 -j2 toolset=clang-7 cxxflags="-fsanitize=undefined -fno-sanitize-recover=undefined" linkflags="-fsanitize=undefined" libs/thread/test//ex_tss -a`

```
testing.capture-output bin.v2/libs/thread/test/ex_tss.test/clang-linux-7/debug/threadapi-pthread/threading-multi/visibility-hidden/ex_tss.run
====== BEGIN OUTPUT ======
libs/thread/src/pthread/thread.cpp:114:37: runtime error: member call on address 0x000001938d10 which does not point to an object of type 'boost::detail::tss_cleanup_function'
0x000001938d10: note: object is of type 'boost::thread_specific_ptr<int>::delete_data'
 00 00 00 00  90 f8 67 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  31 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'boost::thread_specific_ptr<int>::delete_data'

EXIT STATUS: 1
====== END OUTPUT ======
```

**Update**:
  - GCC ticket https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80963 
  - LLVM ticket https://bugs.llvm.org/show_bug.cgi?id=39191

Kojoley:
I have used `shared_ptr` for storing user data directly and custom destructor to handle user supplied destructor and lack of `release` on `shared_ptr`.

This can be further improved by replacing `shared_ptr` with `local_shared_ptr`.

Fixes #236.

### DIFF
--- a/include/boost/thread/pthread/thread_data.hpp
+++ b/include/boost/thread/pthread/thread_data.hpp
@@ -92,16 +92,7 @@ namespace boost
         struct shared_state_base;
         struct tss_cleanup_function;
         struct thread_exit_callback_node;
-        struct tss_data_node
-        {
-            boost::shared_ptr<boost::detail::tss_cleanup_function> func;
-            void* value;
-
-            tss_data_node(boost::shared_ptr<boost::detail::tss_cleanup_function> func_,
-                          void* value_):
-                func(func_),value(value_)
-            {}
-        };
+        typedef boost::shared_ptr<void> tss_data_node;
 
         struct thread_data_base;
         typedef boost::shared_ptr<thread_data_base> thread_data_ptr;
--- a/include/boost/thread/tss.hpp
+++ b/include/boost/thread/tss.hpp
@@ -15,15 +15,9 @@ namespace boost
 {
     namespace detail
     {
-        struct tss_cleanup_function
-        {
-            virtual ~tss_cleanup_function()
-            {}
-
-            virtual void operator()(void* data)=0;
-        };
-
-        BOOST_THREAD_DECL void set_tss_data(void const* key,boost::shared_ptr<tss_cleanup_function> func,void* tss_data,bool cleanup_existing);
+        BOOST_THREAD_DECL void update_tss_data(void const* key,boost::shared_ptr<void> data);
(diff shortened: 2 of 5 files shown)


viboes:
Please, could you describe what is the problem you are trying to fix, how have you fixed it and how do you probe that it has been fixed?

The PR is quite consequent. Do you believe that it could be possible to identify minor issues and minor fixes or refactorings that can be understood better?

Thanks anyway for your interest and working on those PR, really.

Kojoley:
> What do you mean by "This can be further improved by replacing shared_ptr with local_shared_ptr."?

Because it is a thread local storage and `shared_ptr` is designed to be used in multiple threads it uses atomics for counter, by replacing `shared_ptr` with `local_shared_ptr` it will no longer spin atomic counters on c++98.

> Please, could you describe what is the problem you are trying to fix, how have you fixed it and how do you probe that it has been fixed?

There is no actual problem in `thread_specific_ptr` except strange usage of `shared_ptr`, that's why the PR is called `Simplify thread specific storage`, however because previously user variable destructor were implemented with virtual classes the `tss_cleanup_function` were producing virtual tables in every object file (`-Wweak-vtables` clang warning) and also a false positive in UBSan (with `-fvisibility=hidden` which is default now in `b2`). Those are gone with the PR and can be probed.

> The PR is quite consequent. Do you believe that it could be possible to identify minor issues and minor fixes or refactorings that can be understood better?

I am sorry, this PR is very simple and I am not sure how it can be splitted. The only thing  I can undone is replacing `set_tss_data` with three independent function, but it is actually will be a complexification.

Kojoley:
Closed and opened to restart travis

Kojoley:
Nevermind, I found my solution worse than the current (it allocates on every reset), this could be fixed with `shared_ptr` aliasing, but I am not sure I want continue this.

PR #240, opened 2018-10-04
Fixed lockable_traits bugs

### LINKED ISSUE #222: MSVC++17: boost::synchronized_value with boost::synchronize
I'm trying to compile simple code using boost:synchronized_value with boost::synchronize but i'm having problems doing that:

    #include <iostream>
    #include <boost\thread\synchronized_value.hpp>

    int main()
    {
	    boost::synchronized_value<int> val1 = 10;
	    boost::synchronized_value<int> val2 = 20;

	    auto tuple = boost::synchronize(val1, val2);

	    std::cout << *std::get<0>(tuple) << *std::get<1>(tuple) << "\n";

	    return 0;
    }

will result in this error:

    boost::mutex::mutex(const boost::mutex &)" : Es wurde versucht, auf eine gelöschte Funktion zu verweisen
    (use of deleted function)

What i have found out is that this only happens with MSVC++ compilers and only with /std:c++17
if i try to use c++14 it will compile fine, but i want to use c++17 features.
I installed boost with vcpkg if that matters.

Kojoley:
Traits does not handle `noexcept` qualifiers (which are part of function type in C++17) and wrongly did not detect methods.
Because of this on MSVC 14.1 with -std=c++17 `boost::is_mutex_type<boost::mutex>::value` is false and `boost::lock(mutex)` does not compile.

Fixes #222.

### DIFF
--- a/include/boost/thread/lockable_traits.hpp
+++ b/include/boost/thread/lockable_traits.hpp
@@ -11,7 +11,12 @@
 
 #include <boost/assert.hpp>
 #include <boost/detail/workaround.hpp>
+#include <boost/type_traits/integral_constant.hpp>
+#ifdef BOOST_NO_CXX11_SFINAE_EXPR
 #include <boost/type_traits/is_class.hpp>
+#else
+#include <boost/type_traits/declval.hpp>
+#endif
 
 #include <boost/config/abi_prefix.hpp>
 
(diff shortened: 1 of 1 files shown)


Kojoley:
Hm, there is no boost macro to detect P0012R1, I think that it would be better to switch to decltype based detection.

viboes:
I'm wondering to remove those unstable test (depending on the time) from the  regression tests and execute them only on machines I master.

On the other side I Would need to add tests that replace the real time and this needs a lot of work.

PR #241, opened 2018-10-04
Suppress variadic macro warnings

Kojoley:
The workaround was stolen from `boost/static_assert.hpp` and works in Clang too


viboes:
Yeah, why I didn't thought to this before ?

PR #242, opened 2018-10-05
CI: Limit single test execution time to 30 seconds

Kojoley:
It should help identify hanging tests that push CI over build time limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -362,7 +362,7 @@ install:
 script:
   - |-
     echo "using $TOOLSET : : $COMPILER : <cxxflags>-std=$CXXSTD ;" > ~/user-config.jam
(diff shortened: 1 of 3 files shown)


Kojoley:
The timeouts on travis could be avoided either by fixing `execution_monitor` (like this b5743ca4d1746a60a7c4b43d6cff118b4f2f5201) or by increasing timeout, the choice is yours.

viboes:
IIUC the -l30, this will just prevent from spending the maximum time allowed, and limit each test to 30s, isn't it?

Kojoley:
Yes, it is. Did I worded the title badly?
As you can see, Appveyor did not hit the hour limit with this and you can find which tests hang (or too slow).

PR #249, opened 2018-11-17
Simplify TSS cleanup routines. Fixes #236

### LINKED ISSUE #236: UBSan false positive in thread_specific_ptr

Kojoley:
Instead of wrapping a default or user provided destructor into a virtual
class and placing it into a shared_ptr it is now stored directly with
an elided type, to not introduce UB it is not called directly but through
a helper function which casts it back to the original type before calling.

I tried to touch as little lines as possible to make the review simpler.

### DIFF
--- a/include/boost/thread/pthread/thread_data.hpp
+++ b/include/boost/thread/pthread/thread_data.hpp
@@ -94,12 +94,15 @@ namespace boost
         struct thread_exit_callback_node;
         struct tss_data_node
         {
-            boost::shared_ptr<boost::detail::tss_cleanup_function> func;
+            typedef void(*cleanup_func_t)(void*);
+            typedef void(*cleanup_caller_t)(cleanup_func_t, void*);
+
+            cleanup_caller_t caller;
+            cleanup_func_t func;
             void* value;
 
-            tss_data_node(boost::shared_ptr<boost::detail::tss_cleanup_function> func_,
-                          void* value_):
-                func(func_),value(value_)
+            tss_data_node(cleanup_caller_t caller_,cleanup_func_t func_,void* value_):
+                caller(caller_),func(func_),value(value_)
             {}
         };
 
--- a/include/boost/thread/tss.hpp
+++ b/include/boost/thread/tss.hpp
@@ -6,8 +6,8 @@
 // (C) Copyright 2007-8 Anthony Williams
 
 #include <boost/thread/detail/config.hpp>
-#include <boost/shared_ptr.hpp>
-#include <boost/thread/detail/thread_heap_alloc.hpp>
+
+#include <boost/type_traits/add_reference.hpp>
 
 #include <boost/config/abi_prefix.hpp>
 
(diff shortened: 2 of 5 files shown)


viboes:
Thanks for working on this.
I believed that the issue was the the sanitizer itself.
Please, could you describe where the UB reported in #236 was so that I understand better the PR?

@pdimov Please, could you check this and tell us what do you think?

Of course, I couldn't accept this change for boost.1.69, as it is too late.

pdimov:
The original code is a bit odd, I suspect it dates from premodern times when shared_ptr didn't have deleters and boost::function didn't exist. Nowadays you could just use shared_ptr<void>( value, func ), or boost::function.

The patch works too, I suppose. It's a matter of taste which one to pick.

Kojoley:
> The original code is a bit odd, I suspect it dates from premodern times when shared_ptr didn't have deleters and boost::function didn't exist. Nowadays you could just use shared_ptr<void>( value, func ), or boost::function.

The code is now *that* old, it is from 2007, and it was using boost::function  332dd988e461fda0b5d7c038aa3110c61979b8d6.

> @Kojoley Could we summarize that there is no real bug and that the patch is just silencing the sanitizer?

You can treat it like that.

> Are we on UB land?

This case is not documented to be UB, even throwing and catching hidden symbols is not explicitly banned (see boostorg/function#24).

Again:

> There is no consensus. GGC and Clang devs says that it is not a bug in sanitizer.

> Could you clarify what is this fixing aside #236?

Nothing.

viboes:
If this is not fixing nothing clear yet, why do you want I apply this PR then?

Kojoley:
OMG, you are killig me. The PR:

  1. Removes (potential) UB.
  2. Simplifies the code by reducing number of dependencies.
  4. Makes TSS faster by removing `shared_ptr`.

viboes:
I've decided to give it a chance to this PR. We have now the time to fix it if there will be some regressions.

PR #260, opened 2018-12-31
Fix "interruption_point" defined twice

DesWurstes:
```
In file included from /usr/local/include/boost/thread/thread_only.hpp:22,
                 from /usr/local/include/boost/thread/thread.hpp:12,
                 from /usr/local/include/boost/thread.hpp:13,
                 from validation.cpp:48:
/usr/local/include/boost/thread/detail/thread.hpp:591:32: warning: redundant redeclaration of 'void boost::this_thread::interruption_point()' in same scope [-Wredundant-decls]
         void BOOST_THREAD_DECL interruption_point();
                                ^~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/boost/thread/condition_variable.hpp:16,
                 from ./checkqueue.h:13,
                 from validation.cpp:12:
/usr/local/include/boost/thread/pthread/condition_variable.hpp:32:32: note: previous declaration of 'void boost::this_thread::interruption_point()'
         void BOOST_THREAD_DECL interruption_point();
                                ^~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/include/boost/thread/detail/thread.hpp
+++ b/include/boost/thread/detail/thread.hpp
@@ -18,6 +18,9 @@
 #if defined BOOST_THREAD_USES_DATETIME
 #include <boost/thread/xtime.hpp>
 #endif
+#if defined BOOST_THREAD_PROVIDES_INTERRUPTIONS
+#include <boost/thread/interruption.hpp>
+#endif
 #include <boost/thread/detail/thread_heap_alloc.hpp>
 #include <boost/thread/detail/make_tuple_indices.hpp>
 #include <boost/thread/detail/invoke.hpp>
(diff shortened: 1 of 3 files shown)


viboes:
I can not accept this PR.
If the declaration must be only once, then we should avoid having it in two files. 
I would accept a PR where the declaration is in another file included by both :)

DesWurstes:
Thanks for the quick reply! Is it okay now?

include/boost/thread/detail/thread.hpp
```diff
@@ -588,7 +591,6 @@ namespace boost
 #endif
 
 #if defined BOOST_THREAD_PROVIDES_INTERRUPTIONS
-        void BOOST_THREAD_DECL interruption_point();
```
viboes: Wondering if the file shouldn't contains the declarations related to interruptions :)

viboes: In this case the file name should be interrupt.hpp

viboes:
Thanks fro the quick reaction. 

Please let me know what do you think about moving all the interrupt related declarations declarations to a single interrupt.hpp file

DesWurstes:
Is that okay now?

viboes:
Yes, it is ok now :)

PR #262, opened 2019-01-03
Fix MinGW warnings about violation of the strict aliasing rules

Lastique:
Use `memcpy` to convert between long and the bitfield structure and use `memcmp` to compare the two bitfields. Compilers should be smart enough to optimize away these string operations and generate virtually the same code as before.


viboes:
Thaks for the PR.

PR #266, opened 2019-01-14
Remove linking with Boost.System

Lastique:
Since Boost.System is now header-only, no need to link with the library.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -54,7 +54,6 @@ project boost/thread
       #<define>BOOST_SYSTEM_NO_DEPRECATED
       #<define>BOOST_THREAD_DONT_PROVIDE_INTERRUPTIONS
 
-      <library>/boost/system//boost_system
        #-pedantic -ansi -std=gnu++0x -Wextra -fpermissive
         <warnings>all
         <toolset>gcc:<cxxflags>-Wextra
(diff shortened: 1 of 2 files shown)


viboes:
Humm, what happens if a User wants to have Boost.Thread linked with Boost.System?

BTW, Boost.Chrono is also header only, so shouldn't Boost.Thread remove also the link with Boost.Chrono if we apply your logic?

I believe that at the end it is up to the user to select his configuration, isn't it?

I would appreciate a patch of the build system that is able to configure Boost.Thread/Boost.Chrono so that it can be build either linking with Boost.Chrono/Boost.System or not.

viboes:
Oh, I see that Boost.Thread link conditionaly with Boost.Chrono depending on whether we are building with threadapi=pthread or not.

I believe thsi should be independent of the value of threadapi.

How other non-header only libraries are configured in Boost?
Which ones depend on other Boost libraries?
Is there something that it is agreed on in Boost?

Lastique:
Maybe you misunderstood me. Boost.System is unconditionally header-only now, there's no config option to make it otherwise. Its compiled library is empty and it only exists to avoid breaking dependent libraries and applications. Given this, why would anyone want to link with it or have a config option in Boost.Thread? What is your reason in keeping the linking dependency?

I'm not sure what's the situation with Boost.Chrono, this PR is not about it. However, IMHO, if Boost.Chrono can be (or is) header-only without sacrificing its features or other major compromises, I don't see why it shouldn't be.

viboes:
Sorry, I was not aware that Boost.System had nothing inside the binary library now.
This change completely the context.

PR #268, opened 2019-01-19
Add self contained header tests and fix discovered bugs

Lastique:
Added tests for self-contained headers.

The new set of tests iterates over Boost.Thread public headers and verifies that
each header is self-contained, i.e. doesn't have any missing includes and
contains syntactically correct content. On Windows and Cygwin, a second test
per each header is generated, which includes the header after windows.h. This
verifies that the header doesn't have conflicts with Windows SDK and that
including windows.h does not break platform detection in Boost.Thread.

This set of tests would have detected the bug fixed by https://github.com/boostorg/thread/pull/263.

Fixed bugs discovered by the standalone header tests.

- Added missing includes.
- Added missing namespace qualification for Boost.Chrono casts.
- Added typedefs renaming polymorphic lockable wrappers to a more consistent
  naming scheme. This fixes incorrect name references.
- Fixed incorrect duration casts in polymorphic lockable wrappers.
- Renamed upgrade_lockable_adapter to poly_upgrade_lockable_adapter, since
  upgrade_lockable_adapter already exists and has a different meaning.
- Use BOOST_THREAD_FUTURE instead of future in executors and task region.
- Disable serial_executor_cont when executors or future continuations
  are not enabled.
- Marked sync_queue_is_closed exception as visible.

### DIFF
--- a/include/boost/thread/executors/serial_executor_cont.hpp
+++ b/include/boost/thread/executors/serial_executor_cont.hpp
@@ -10,13 +10,20 @@
 #define BOOST_THREAD_SERIAL_EXECUTOR_CONT_HPP
 
 #include <boost/thread/detail/config.hpp>
+
+#if defined(BOOST_THREAD_PROVIDES_FUTURE_CONTINUATION) && defined(BOOST_THREAD_PROVIDES_EXECUTORS)
+
+#include <exception> // std::terminate
+#include <boost/throw_exception.hpp>
 #include <boost/thread/detail/delete.hpp>
 #include <boost/thread/detail/move.hpp>
 #include <boost/thread/concurrent_queues/sync_queue.hpp>
 #include <boost/thread/executors/work.hpp>
 #include <boost/thread/executors/generic_executor_ref.hpp>
+#include <boost/thread/mutex.hpp>
 #include <boost/thread/future.hpp>
 #include <boost/thread/scoped_thread.hpp>
+#include <boost/thread/lock_guard.hpp>
 
 #include <boost/config/abi_prefix.hpp>
 
@@ -32,7 +39,7 @@ namespace executors
   private:
 
     generic_executor_ref ex_;
-    future<void> fut_; // protected by mtx_
+    BOOST_THREAD_FUTURE<void> fut_; // protected by mtx_
     bool closed_; // protected by mtx_
     mutex mtx_;
 
@@ -44,7 +51,7 @@ namespace executors
       };
       continuation(BOOST_THREAD_RV_REF(work) tsk)
       : task(boost::move(tsk)) {}
-      void operator()(future<void> f)
+      void operator()(BOOST_THREAD_FUTURE<void> f)
       {
         try {
           task();
@@ -166,5 +173,7 @@ using executors::serial_executor_cont;
 }
 
 #include <boost/config/abi_suffix.hpp>
+
+#endif // defined(BOOST_THREAD_PROVIDES_FUTURE_CONTINUATION) && defined(BOOST_THREAD_PROVIDES_EXECUTORS)
 
 #endif
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
   include:
     - os: linux
       compiler: g++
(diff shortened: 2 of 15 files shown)


include/boost/thread/executors/serial_executor_cont.hpp
```diff
@@ -32,7 +39,7 @@ namespace executors
   private:
 
     generic_executor_ref ex_;
-    future<void> fut_; // protected by mtx_
+    BOOST_THREAD_FUTURE<void> fut_; // protected by mtx_
```
viboes: This  is not absolutely necessary as this header is supported only after version 3.
Anyway, it doesn't hurt.

viboes:
Thanks for providing those test.

include/boost/thread/executors/serial_executor_cont.hpp
Lastique: I did not see this documented anywhere, nor guarded by any macro checks. There must be a macro check so that the header can be included by the tests.

viboes:
Maybe it was not explicitly documented.

But see trying to use executors without version 3 seems to not work
See
https://api.travis-ci.org/v3/job/481848302/log.txt

Lastique:
If you compile all headers with all versions, the tests will fail and thus will be of no use. If you only want to compile headers with the version numbers they support then I'm not sure how to do that since the tests are generated automatically and have no knowledge of the supported versions. Therefore I think all headers must be compatible with all versions, even if it means the headers are empty for unsupported versions, and compiling for the default version (which should be the latest one) is enough.

Also note that the tests already take a long time to compile, which adds to the already long time to complete other tests. Compiling all headers for every version will increase this time N-fold.

include/boost/thread/executors/serial_executor_cont.hpp
viboes: Unfortunately BOOST_THREAD_PROVIDES_EXECUTORS is not defined in any BOOST_THREAD_VERSION
```
Feature                             | Anti-Feature                     | V2   | V3  | V4
PROVIDES_EXECUTORS | -                                         | NO  | NO | NO
PROVIDES_FUTURE        | DONT_PROVIDE_FUTURE | NO | YES | YES
```

https://www.boost.org/doc/libs/master/doc/html/thread/build.html#thread.build.configuration

Lastique: Yes, I noticed that.

viboes:
I don't see why ant test should fails. If each file is protected with respect to the good enablers, non should fail as you have already done for one of the files.
So yes, all headers must be compatible with all version. This is what I was asking.
I know that the test will increase the time. I would like to have the test available on demand at least.

Lastique:
Ok. But just to be clear, I don't know what header requires what version, so I'm not planning to add preprocessor checks. I will say though that I'm most interested for the tests to pass in the default version, as this is likely the version most users will use.

viboes:
I'm saying the same as you: "Therefore I think all headers must be compatible with all versions, even if it means the headers are empty for unsupported versions."

However, compiling only for the default version will not check the files are self contained :(

I don't know which versions the people are using, but the default version is quite limited in features.

viboes:
I'm preparing a fix for some of the issues still present on the self consistent files.

Lastique:
> I don't know which versions the people are using, but the default version is quite limited in features.

Well, we're using the default version in other Boost libraries. I'm not aware of anyone setting a non-default version.

If other versions are more useful, why not make it the default?

> I'm preparing a fix for some of the issues still present on the self consistent files.

Ok, good, thanks.

viboes:
> If other versions are more useful, why not make it the default?

backward compatibility :(

Lastique:
Those who want a specific API version could (and should) request it explicitly in their project. The default should be the version you recommend people to use in new projects or the projects that are willing to update to a new version.

viboes:
I proposed long time ago a plan to move to new versions and the people rejected my plan. This is why we have yet so much versions in Boost.Thread.

I did something wrong and I'm tired now to make it better. 

If someone wants to make a really new version on separated folders, please go ahead.

If someone want to propose to change the default version for Boost.Thread, please go ahead and propose it on the Boost development mailing list.

Issue #271, opened 2019-03-17
boost::executors::scheduler not work as expected in 1.69


DengZuoheng:
In following code, every 500ms submit a work, but each work's time duration less than before:
~~~
#include <boost/chrono/system_clocks.hpp>
#include <boost/thread/executors/scheduler.hpp>

int main() {
	boost::executors::scheduler<boost::chrono::steady_clock> sc;
	auto start = boost::chrono::steady_clock::now();
	int count = 10;
	for (int i = 0; i < count; ++i) {
		sc.submit_after([=] {
			std::cout << count - i << "s" << std::endl;
			auto dur = boost::chrono::steady_clock::now() - start;
			std::cout << boost::chrono::duration_cast<boost::chrono::milliseconds>(dur) << std::endl;
		}, boost::chrono::seconds(count - i));
		boost::this_thread::sleep_for(boost::chrono::milliseconds(500));
		std::cout << i << std::endl;
	}

	boost::this_thread::sleep_for(boost::chrono::seconds(count * 2));
	return 0;

}
~~~

1.69 behave different with 1.66, while 1.66's output is:

~~~
0
1
2
3
4
5
6
7
8
9
1s
5507 milliseconds
2s
6006 milliseconds
3s
6506 milliseconds
4s
7004 milliseconds
5s
7504 milliseconds
6s
8004 milliseconds
7s
8504 milliseconds
8s
9003 milliseconds
9s
9501 milliseconds
10s
10001 milliseconds
~~~

but 1.69's output is 

~~~
0
1
2
3
4
5
6
7
8
9
1s
10001 milliseconds
2s
10002 milliseconds
3s
10002 milliseconds
4s
10003 milliseconds
5s
10003 milliseconds
6s
10004 milliseconds
7s
10005 milliseconds
8s
10005 milliseconds
9s
10006 milliseconds
10s
10006 milliseconds
~~~

1.66 is what I want. It's that a bug of 1.69 or I use it wrong?

Thanks.

viboes:
```
I confirm you figures.

https://wandbox.org/permlink/Z3oiCHlYTsMQdOk6

The difference is between 1.66 and 1.67 and my understanding is as well that the result for 1.66 are better.

However, the results for 1.67 and 1.69 are also good. This is because the request is to execute after the specified duration and the exeution take place after this duration is elapsed.

I believe the changes is due to 
https://github.com/boostorg/thread/issues/162 fix as much time-related issues as possible and improve the QOI

[...]

PR #277, opened 2019-04-01
Fix sync_timed_queue time jump issues

austin-beer:
* Added missing sync_timed_queue::pull_for() time jump tests.
* Added pull_until() and pull_for() to the test for #271.
* Fixed time jump issues that were re-introduced while fixing #271.
* Fixed time jump issues with sync_timed_queue::pull_for().

### DIFF
--- a/include/boost/thread/concurrent_queues/sync_timed_queue.hpp
+++ b/include/boost/thread/concurrent_queues/sync_timed_queue.hpp
@@ -16,6 +16,8 @@
 #include <boost/chrono/system_clocks.hpp>
 #include <boost/chrono/chrono_io.hpp>
 
+#include <algorithm> // std::min
+
 #include <boost/config/abi_prefix.hpp>
 
 namespace boost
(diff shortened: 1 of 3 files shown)


include/boost/thread/concurrent_queues/sync_timed_queue.hpp
```diff
@@ -88,8 +88,8 @@ namespace detail
     T pull();
     void pull(T& elem);
 
-    template <class WClock, class Duration>
-    queue_op_status pull_until(chrono::time_point<WClock,Duration> const& tp, T& elem);
+    template <class Duration>
+    queue_op_status pull_until(chrono::time_point<clock,Duration> const& tp, T& elem);
```
viboes: I'm not saying the implementation was correct, but why you don't want to support other clocks?

include/boost/thread/concurrent_queues/sync_timed_queue.hpp
```diff
@@ -122,8 +122,7 @@ namespace detail
     inline bool not_empty_and_time_reached(lock_guard<mutex>& lk) const;
 
     bool wait_to_pull(unique_lock<mutex>&);
-    template <class WClock, class Duration>
-    queue_op_status wait_to_pull_until(unique_lock<mutex>&, chrono::time_point<WClock, Duration> const& tp);
+    queue_op_status wait_to_pull_until(unique_lock<mutex>&, TimePoint const& tp);
```
viboes: Why do you want to be more generic here? Do you have other timepoints other than boost::chrono::timepoint? Are you thinking on std::chrono::timepoint?

include/boost/thread/concurrent_queues/sync_timed_queue.hpp
austin-beer: Currently, if you create a sync_timed_queue<chrono::steady_clock> and then call pull_until() and pass it a chrono::system_clock (or any other clock) time point, the code will fail to compile because it tries to compare `tp` against `clock::now()` in wait_to_pull_until(). So we have two options:

1. Add code to convert between clocks in wait_to_pull_until(), which adds complexity to the code.
2. Change pull_until() to make it clear that mixing clocks is not supported.

I chose number 2 because that's the existing behavior and because it keeps the implementation simpler and because I can't think of a good use-case where someone would need to mix clocks. What do you think?

include/boost/thread/concurrent_queues/sync_timed_queue.hpp
austin-beer: Since `wait_to_pull_until()` is an internal/private function, it simplified the code to have the type of `tp` match the type of `super::data_.top().time`, which is also of type `TimePoint`. Then in `pull_until()` I use `time_point_cast()` to convert from the user's time point type to `TimePoint`.

include/boost/thread/concurrent_queues/sync_timed_queue.hpp
viboes: Lets go with approach 2 by now. I will see if I can find out a valid use case.

include/boost/thread/concurrent_queues/sync_timed_queue.hpp
viboes: Okay, I see.

Issue #283, opened 2019-05-17
can't build on solaris 11.3 using gcc 8.3.0


ohhmm:
https://github.com/boostorg/boost/issues/291

https://github.com/boostorg/boost/files/3190241/log.txt

./boost/thread/pthread/thread_data.hpp:60:5: error: missing binary operator before token "_sysconf"

viboes:
Please, could you check why do we have this issue

```
./boost/thread/pthread/thread_data.hpp:60:5: error: missing binary operator before token "_sysconf"
 #if PTHREAD_STACK_MIN > 0
```

is PTHREAD_STACK_MIN defined as _sysconf on this platform?

Could you provide the preprocessed text?

viboes:
Any PR with a reasonable solution is welcome. I will not do anything from my side as I don't have access to this platform.

johnparke:
I had the same problem on SunOS 5.10, gcc version 7.4.0.

The problem is caused by the statement following the sysconf reference, the #if statement:

#ifdef BOOST_THREAD_USES_GETPAGESIZE
          std::size_t page_size = getpagesize();
#else
          std::size_t page_size = ::sysconf( _SC_PAGESIZE);
#endif
#if PTHREAD_STACK_MIN > 0
          if (size<PTHREAD_STACK_MIN) size=PTHREAD_STACK_MIN;
#endif

In the Solaris limits.h file, PTHREAD_STACK_MIN is defined as:

`#define	PTHREAD_STACK_MIN	((size_t)_sysconf(_SC_THREAD_STACK_MIN))`

So the if statement creates the error condition. I think what is expected is a value such as 16384.

datalogics-robb:
This issue was introduced with this commit:

https://github.com/boostorg/thread/commit/e54f7e39606b807b8f9fc84fbd58cafeabea6562

The comment says that it was to fix a warning "comparing unsigned to zero" so I'm sure this would be safe to revert.

https://github.com/boostorg/thread/pull/297

MarcoFalke:
The issue is fixed and can be closed?

seerviashish:
Still facing this issue. Is the issue closed? Ubuntu 20.04 and g++ --version = 10.3.0

johnparke:
Not using Solaris anymore, so don’t need to build on that platform.

 

Thanks,

 

John

hilga007:
> Still facing this issue. Is the issue closed? Ubuntu 20.04 and g++ --version = 10.3.0

[...]

PR #297, opened 2019-11-20
Revert change to elide a warning that caused Solaris builds to fail.

datalogics-robb:
Revert a minor change to elide a warning regarding comparison of unsigned to zero that caused solaris platforms to fail to build.

https://github.com/boostorg/thread/issues/283


datalogics-robb:
There is a single CI failure, 388.10 - but that doesn't look like it could be related to the change in this PR.

PR #305, opened 2020-01-22
By default #define BOOST_THREAD_USE_LIB in Windows for clang

dmenendez-gruposantander:
I was trying to compile my code in Windows with clang-cl and BOOST_ALL_NO_LIB and I was getting weird linker errors about missing *imported* symbols:  undefined symbol: __declspec(dllimport) public: void __thiscall boost::thread::join(void)

This led me to find out that clang-cl was a second class citizen wrt the definition of BOOST_THREAD_USE_LIB by default in Windows.

I propose to add clang to the compilers that get BOOST_THREAD_USE_LIB defined by default in Windows.


dmenendez-gruposantander:
I see PR #290 already took care of this, but I was confused because I didn't see it merged in the 1.72.0 release

PR #309, opened 2020-03-03
Cleanup header includes

### LINKED ISSUE #307: boost/thread/detail/thread.hpp #includes boost/bind.hpp unconditionally but uses it conditionally
I can see boost::bind() used only if BOOST_THREAD_PROVIDES_VARIADIC_THREAD is *not* defined

Could the #include <boost/bind.hpp> be moved inside a #ifndef BOOST_THREAD_PROVIDES_VARIADIC_THREAD section? In fact I'd move it inside an #else section under line 50

Boost.Bind brings in a lot of cruft, that is, increased compile times and funny warnings `boost/bind/placeholders.hpp(54): note: see reference to class template instantiation 'boost::arg<9>' being compiled`

Happy to provide a PR

Lastique:
1. Make inclusion of `boost/bind/bind.hpp` conditional in some cases, when the code actually conditionally uses `boost::bind`. Reduces compile-time overhead and fixes https://github.com/boostorg/thread/issues/307.

2. Remove some unnecessary uses of `boost::ref`. This allows to avoid including `boost/core/ref.hpp` in a few places, and avoids the associated template instantiation overhead in others.

3. Replace deprecated header includes with the more recent alternatives. For example: `boost/detail/lightweight_test.hpp` -> `boost/core/lightweight_test.hpp`, `boost/ref.hpp` -> `boost/core/ref.hpp`.

4. Replace some blanket includes with the more fine-grained ones. For example, `boost/utility.hpp`, `boost/atomic.hpp`. This reduces compile time overhead.

5. Add some missing includes, for example, `boost/core/ref.hpp` and `boost/type_traits/is_same.hpp`.

6. Replace uses of `std::is_same` with `boost::is_same` (with the corresponding included header) since the standard type_traits header presence and validity is not tested by the code. Using `boost::is_same` makes the code more portable.

### DIFF
--- a/include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
+++ b/include/boost/thread/concurrent_queues/detail/sync_deque_base.hpp
@@ -12,6 +12,7 @@
 //////////////////////////////////////////////////////////////////////////////
 
 #include <boost/bind/bind.hpp>
+#include <boost/core/ref.hpp>
 
 #include <boost/thread/detail/config.hpp>
 #include <boost/thread/condition_variable.hpp>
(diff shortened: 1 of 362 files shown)


Lastique:
@viboes @pdimov A kind reminder. Could we merge this to develop and master before 1.73?

pdimov:
This PR has too many unrelated changes so I didn't feel comfortable applying it. Instead, I just changed `<boost/bind.hpp>` to `<boost/bind/bind.hpp>` in c13beec81ce4b67f38359ebf6db8c2f8f02c3d62.

Lastique:
@pdimov Thanks. I will update this PR accordingly.

PR #310, opened 2020-03-03
Avoid relying on implicit copy constructor/operator deprecated in C++11

Lastique:
C++11 deprecates implicit default copy constructors and operators if the class has user-defined destructor or copy constructor/operator. gcc 9 generates warnings when this deprecated language feature is used. This commit fixes that by providing user-defained copy constructors/operators where needed. The added definitions are equivalent to the implicitly generated by the compiler.

Additionally, for `thread::id` added move constructor/operator when rvalue references are available. Move is more efficient when the id uses a smart pointer internally.

### DIFF
--- a/include/boost/thread/detail/thread.hpp
+++ b/include/boost/thread/detail/thread.hpp
@@ -637,10 +637,6 @@ namespace boost
 #endif
         {}
 
-        id(const id& other) BOOST_NOEXCEPT :
-            thread_data(other.thread_data)
-        {}
-
         bool operator==(const id& y) const BOOST_NOEXCEPT
         {
             return thread_data==y.thread_data;
--- a/test/sync/futures/promise/set_value_const_pass.cpp
+++ b/test/sync/futures/promise/set_value_const_pass.cpp
@@ -23,6 +23,7 @@
 #include <boost/thread/future.hpp>
 #include <boost/detail/lightweight_test.hpp>
 #include <boost/static_assert.hpp>
+#include <boost/config.hpp>
 
 #ifdef BOOST_MSVC
 # pragma warning(disable: 4702) // unreachable code
@@ -37,6 +38,9 @@ struct A
   {
     throw 10;
   }
+#if !defined(BOOST_NO_CXX11_DEFAULTED_FUNCTIONS)
+  A& operator= (const A&) = default;
+#endif
 };
 
 int main()
--- a/example/thread_guard.cpp
+++ b/example/thread_guard.cpp
@@ -20,6 +20,7 @@ struct func
     int& i;
 
     func(int& i_):i(i_){}
+    func(func const& other):i(other.i){}
 
     void operator()()
     {
--- a/test/shared_mutex_locking_thread.hpp
+++ b/test/shared_mutex_locking_thread.hpp
@@ -7,6 +7,7 @@
 //  accompanying file LICENSE_1_0.txt or copy at
 //  http://www.boost.org/LICENSE_1_0.txt)
 
+#include <boost/config.hpp>
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/condition_variable.hpp>
 #include <boost/thread/shared_mutex.hpp>
@@ -37,6 +38,10 @@ public:
         unblocked_count_mutex(unblocked_count_mutex_),
         finish_mutex(finish_mutex_)
     {}
+
+#if !defined(BOOST_NO_CXX11_DEFAULTED_FUNCTIONS)
+    locking_thread(locking_thread const&) = default;
+#endif
 
     void operator()()
     {
@@ -84,6 +89,10 @@ public:
         unblocked_mutex(unblocked_mutex_),unblocked_count(unblocked_count_)
     {}
 
+#if !defined(BOOST_NO_CXX11_DEFAULTED_FUNCTIONS)
+    simple_writing_thread(simple_writing_thread const&) = default;
+#endif
+
     void operator()()
     {
         boost::unique_lock<boost::shared_mutex>  lk(rwm);
@@ -114,6 +123,10 @@ public:
         rwm(rwm_),finish_mutex(finish_mutex_),
         unblocked_mutex(unblocked_mutex_),unblocked_count(unblocked_count_)
     {}
+
+#if !defined(BOOST_NO_CXX11_DEFAULTED_FUNCTIONS)
+    simple_reading_thread(simple_reading_thread const&) = default;
+#endif
 
     void operator()()
     {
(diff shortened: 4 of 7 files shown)


include/boost/thread/detail/thread.hpp
```diff
@@ -641,6 +641,27 @@ namespace boost
...
+        id& operator=(id&& other) BOOST_NOEXCEPT
+        {
+            thread_data = static_cast< data&& >(other.thread_data);
+#if defined(BOOST_THREAD_PROVIDES_BASIC_THREAD_ID)
+            other.thread_data = 0;
+#endif
+            return *this;
+        }
+#endif // !defined(BOOST_NO_CXX11_RVALUE_REFERENCES)
+
```
Kojoley, line 664: Can't we just remove the user defined copy constructor instead (it seems to be not doing anything special)

Lastique: The move assignment does have to do something special (assign zero to the move source), which means other operators/constructors are also needed.

Which makes me notice that the move constructor should set the move source to zero. I'll update the PR.

pdimov:
Enumerating members by hand is too error prone for my taste; how about using =default instead?

I'd also split the thread::id change into its own PR. (It seems to break atomic<thread::id>.)

Lastique:
> Enumerating members by hand is too error prone for my taste; how about using =default instead?

That would require C++11, and I wouldn't want to raise the minimum required C++ version for Boost.Thread. Using it conditionally is not useful as I would still have to provide the C++03 version.

> I'd also split the thread::id change into its own PR.

It's related, because `id` has the same problem - it has a copy constructor but not `operator=`.

> (It seems to break `atomic<thread::id>`.)

I believe, it never was valid because `id` was not trivially copyable even before this change. It could be made such, if `id::data` permits (which it doesn't sometimes), but that would be for a different PR.

pdimov:
The implicit members are only deprecated on C++11, so =defaulting them on C++11 doesn't raise a minimum requirement.

pdimov:
Yes, `id` has a manually declared copy constructor. I don't know why this test used to pass.

pdimov:
> but that would be for a different PR.

I'm not sure about that; applying the =default change will automatically make it trivial.

Lastique:
Ok, I can change the tests to use `=default` copy constructors conditionally. Although personally, I like the current code more as it is not conditional.

I'm hesitant about `id` because some (older) compilers deduce `noexcept` incorrectly for defaulted functions, and move assignment/constructor must be `noexcept`. I think, older gcc and Intel had this problem. We don't have a Boost.Config macro to detect these buggy compilers.

Lastique:
Also, the move constructor/assignment are not default anyway, as they need to assign zero to the move source.

pdimov:
There is no such requirement. It's normal and expected for trivial types to perform a copy on move. But the move operations of id, and the discussion about them, really belong in a separate PR.

Lastique:
Ok, I've updated the PR:

- Changed the tests to use defaulted constructors conditionally.
- Removed move support from `thread::id`.
- Left `thread::id` non-trivially copyable because of the buggy compilers wrt. `noexcept` deduction.

Also, the test failure is because `testable_mutex` incorrectly uses `atomic<thread::id>`. Boost.Atomic was modified not long ago to issue a static assert that the type is trivially copyable. The requirement has always been there, it's just it was not enforced before.

`testable_mutex` needs to be fixed in a separate PR. Probably, by using the underlying thread handle type or `thread_data*` in the atomic.

Lastique:
> There is no such requirement. It's normal and expected for trivial types to perform a copy on move.

Yes, but `id` is not trivial. And when a smart pointer is used internally the move source is left in the default-constructed state (i.e. equal to `id()`). I think, the behavior should be consistent regardless of the internal representation of `id`.

test/sync/futures/promise/set_value_const_pass.cpp
```diff
@@ -37,6 +37,9 @@ struct A
   {
     throw 10;
   }
+  A& operator= (const A&)
+  {
```
pdimov, line 41: return *this, but =default is better

include/boost/thread/detail/thread.hpp
```diff
@@ -641,6 +641,12 @@ namespace boost
             thread_data(other.thread_data)
         {}
 
+        id& operator=(const id& other) BOOST_NOEXCEPT
```
pdimov, line 644: =default, and same for the copy constructor

test/sync/futures/promise/set_value_const_pass.cpp
Lastique: Done.

include/boost/thread/detail/thread.hpp
Lastique: As noted before, I disagree, because this will affect `noexcept` for some compilers.

pdimov: How relevant is that in practice? A non-noexcept move constructor causes the copy constructor to be used instead, which may be a pessimization (except of course in the case where the move constructor is more expensive, such as when it clears the rhs unnecessarily); but a non-noexcept copy constructor doesn't do anything.

Lastique: It allows `std::swap` to be `noexcept`, for example. It also affects `noexcept` of any default-generated functions in the user's classes that have `id` as a member. Not to mention that `noexcept` is part of the contract. In general, I consider `noexcept` property more important than `=default`.

pdimov: No compiler on Godbolt mis-noexcepts the =default copy. https://godbolt.org/z/esZKLa

Lastique: Hmm, my experience is quite different: https://github.com/boostorg/atomic/blob/dd4cb7e5400b6f69f4fa5dd909146ad170a2a7d5/include/boost/atomic/detail/config.hpp#L94. I'm not sure why godbolt works. Perhaps, older compilers are unstable and the deduction works in some cases and not in other.

pdimov: I think that your configuration macro refers to the case
```
id( id const& ) noexcept = default;
```
which is indeed not supported by some early versions.

The case
```
id( id const& ) = default;
```
is universally supported, and the only case in which noexcept is not deduced correctly (on Godbolt) is on Intel 13, when there is a nontrivial member with a noexcept copy. https://godbolt.org/z/LhTy_A

We can ifdef this out in principle (this only affects the smart pointer case), but how important is Intel 13 nowadays? 16 and higher all work.

Lastique: I decided to just remove the copy constructor to allow the compiler generate the full set of constructors/assignment operators. Hopefully, it will do the right thing.

Lastique:
I've updated the PR:

- Rebased on current develop.
- Removed copy constructor of `thread::id` to allow all constructors/assignment to be generated, including move.

Lastique:
Thanks, Peter.

PR #312, opened 2020-04-04
Add support for move constructor/assignment to thread::id

Lastique:
Move is more efficient than copy in case if `thread::id` uses a smart pointer internally.

This PR was extracted from https://github.com/boostorg/thread/pull/310.


Lastique:
This should no longer be needed since https://github.com/boostorg/thread/pull/310 was merged.

PR #320, opened 2020-06-14
Fixed optimized away hooks. Fixes #316

### LINKED ISSUE #316: VS 2019: thread_specific_ptr causes unbalanced destructor calls of hold objects
We are using Boost since years in our company with various compilers, among them the Visual Studio compiler family. We are currently considering to use VS 2019 in production code and I'm evaluating the most recent version with the most recent Boost release version 1.72 (I also tested 1.70 which we currently use in production code, and there is no difference for the here discussed issue between both boost versions). Our currently production-use compiler is VS 2017 and the here described issue does not exist. In our code base we have a very simple test case that verifies that the following remark in [https://www.boost.org/doc/libs/1_72_0/doc/html/thread/thread_local_storage.html](https://www.boost.org/doc/libs/1_72_0/doc/html/thread/thread_local_storage.html):

> Note: on some platforms, cleanup of thread-specific data is not performed for 
> threads created with the platform's native API. On those platforms such cleanup 
> is only done for threads that are started with boost::thread unless 
> boost::on_thread_exit() is called manually from that thread.

Our assumption is that on windows and on the linux versions that we support there is no need for calling `boost::on_thread_exit()` manually in threads _not_ started by `boost::thread`, therefore our minimalistic test to validate this, see the following attachment where I have removed all parts that are not related to Windows:

[test_tss_direct.cpp.txt](https://github.com/boostorg/thread/files/4497139/test_tss_direct.cpp.txt)

This test succeeds for VS 2012 and VS 2017, but starts failing with VS 2019: Here the test output is:

`[doTestBoostThreadSpecificPtrCleanup] TSS instance counter is different from 1: 2 `

and the program returns 1 instead of 0. In other words: Only one destructor of the two created `InstanceCountingClass` objects is called. 

_The problem could only be observed for Release builds, but for both 32 and 64 bit_.

For this isolated test I build the required boost libraries using this command (The following restricts for the 32-bit case):

`b2 -j%NUMBER_OF_PROCESSORS% toolset=msvc-14.2 --with-thread --with-atomic address-model=32 link=static,shared variant=release optimization=speed cxxflags="/Zc:inline" --stagedir=%BOOST_HOME%\boost_install\1_72\vc142-test\x86 --compiler="C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\bin\HostX64\x64\cl.exe"`

using the most recent VS 2019 community edition (16.5.4). I created an empty project without changing any of the default compiler settings accept that I added to the Additional Include directories the value "$(BOOST_HOME)\boost_1_72_0;" and to the Additional Linker Directories the value "$(BOOST_HOME)\boost_install\1_72\vc142-test\$(PlatformTarget)\lib;"

This is the list of all effective compiler options as seen from the Command Line All options view:

`/permissive- /GS /GL /analyze- /W3 /Gy /Zc:wchar_t /I"C:\Projects\ThirdParty\boost_1_72_0" /Zi /Gm- /O2 /sdl /Fd"Release\vc142.pdb" /Zc:inline /fp:precise /D "NDEBUG" /D "_CONSOLE" /D "_UNICODE" /D "UNICODE" /errorReport:prompt /WX- /Zc:forScope /Gd /Oy- /Oi /MD /FC /Fa"Release\" /EHsc /nologo /Fo"Release\" /Fp"Release\tss-direct-test.pch" /diagnostics:column `

I should point out that we observed a similar test failure around 2017 with an older boost version but it vanished (reliably) at some point where I either updated the boost version or got a compiler update of VS 2017. Unfortunately we never understood which of these changes had the healing effect in the past, but we were happy and kept this test and run basically once in the week without any test failure since 2017 - up to the point were I started evaluating the VS 2019 compiler. It is completely possible that this is a compiler bug, but before I report it to Microsoft I would like to ask you to have a look at it, because we use all boost libraries (including the thread library) as static libraries and AFAIK the code in `tss_pe.cpp` may be very sensitive to changes in intrinsics.

Thanks for your support in advance!

P.S.: Today I realized that while I made a bunch of experiments to reduce the boost build parameter set to a minimal one I forgot in one of my last tests to delete the bin.v2 directory where cached build results where present. I checked again and found that the issue is reproducible, when the /Zc:inline compiler flag is active _while building boost-thread_ (but not when using /Zc:inline in the depending code). I corrected therefore the reproducer build parameter list above.

Question: Is it possible to perform a complete boost build using the b2 commands but handling *one* library (In this case Boost Thread) special?

Kojoley:
MSVC learned to not emit unreferenced symbols with internal linkage and the
hooks were defined in unnamed namespace which forces internal linkage, even if
you mark a variable `extern`.

Since Boost does not have a stable ABI, does not mangle the namespace with
the version, and the hooks are in `boost` namespace (`boost::on_*`) -- there is
no point in trying to hide some symbols because mixing different versions of
boost static libraries will not work already.

I also renamed the `__xl_ca` variable for consistency and because using double
underscored identifiers is forbidden. (`[lex.name]/3`)

The `extern const` is for verbosity and because they are indeed const (it is
done via pragma already).

### DIFF
--- a/src/win32/tss_pe.cpp
+++ b/src/win32/tss_pe.cpp
@@ -152,19 +152,25 @@ extern BOOL (WINAPI * const _pDefaultRawDllMainOrig)(HINSTANCE, DWORD, LPVOID) =
         static PVAPI_V on_process_init();
         static PVAPI_V on_process_term();
         static void NTAPI on_tls_callback(HINSTANCE, DWORD, PVOID);
+    }
 
+    namespace boost
+    {
         //The .CRT$Xxx information is taken from Codeguru:
         //http://www.codeguru.com/Cpp/misc/misc/threadsprocesses/article.php/c6945__2/
+
+        // Variables below are not referenced anywhere and
+        // to not be optimized away has to have external linkage
 
 #if (_MSC_VER >= 1400)
 #pragma section(".CRT$XIU",long,read)
 #pragma section(".CRT$XCU",long,read)
 #pragma section(".CRT$XTU",long,read)
 #pragma section(".CRT$XLC",long,read)
-        __declspec(allocate(".CRT$XLC")) _TLSCB __xl_ca=on_tls_callback;
-        __declspec(allocate(".CRT$XIU"))_PIFV_ p_tls_prepare = on_tls_prepare;
-        __declspec(allocate(".CRT$XCU"))_PVFV_ p_process_init = on_process_init;
-        __declspec(allocate(".CRT$XTU"))_PVFV_ p_process_term = on_process_term;
+        extern const __declspec(allocate(".CRT$XLC")) _TLSCB p_tls_callback = on_tls_callback;
+        extern const __declspec(allocate(".CRT$XIU")) _PIFV_ p_tls_prepare = on_tls_prepare;
+        extern const __declspec(allocate(".CRT$XCU")) _PVFV_ p_process_init = on_process_init;
+        extern const __declspec(allocate(".CRT$XTU")) _PVFV_ p_process_term = on_process_term;
 #else
         #if (_MSC_VER >= 1300) // 1300 == VC++ 7.0
         #   pragma data_seg(push, old_seg)
@@ -176,30 +182,33 @@ extern BOOL (WINAPI * const _pDefaultRawDllMainOrig)(HINSTANCE, DWORD, LPVOID) =
             //this could be changed easily if required.
 
             #pragma data_seg(".CRT$XIU")
-            static _PIFV_ p_tls_prepare = on_tls_prepare;
+            extern const _PIFV_ p_tls_prepare = on_tls_prepare;
             #pragma data_seg()
 
             //Callback after all global ctors.
 
             #pragma data_seg(".CRT$XCU")
-            static _PVFV_ p_process_init = on_process_init;
+            extern const _PVFV_ p_process_init = on_process_init;
             #pragma data_seg()
 
             //Callback for tls notifications.
 
             #pragma data_seg(".CRT$XLB")
-            _TLSCB p_thread_callback = on_tls_callback;
+            extern const _TLSCB p_thread_callback = on_tls_callback;
             #pragma data_seg()
             //Callback for termination.
 
             #pragma data_seg(".CRT$XTU")
-            static _PVFV_ p_process_term = on_process_term;
+            extern const _PVFV_ p_process_term = on_process_term;
             #pragma data_seg()
         #if (_MSC_VER >= 1300) // 1300 == VC++ 7.0
         #   pragma data_seg(pop, old_seg)
         #endif
 #endif
+    } // namespace boost
 
+    namespace
+    {
 #ifdef BOOST_MSVC
 #pragma warning(push)
 #pragma warning(disable:4189)


Kojoley:
I did not add a test because `tss_test` is already catching the issue. I also checked that this fixes Clang build which had the issue a long before.

Lastique:
@pdimov Since @viboes is not responding, maybe you could merge this fix?

pdimov:
OK, done. I'm getting crashes in some `future` tests on `clang-win`, but that's a preexisting problem. Maybe @Kojoley knows why.

Lastique:
Thanks, Peter. Please, remember to merge to master as well.

pdimov:
I will.

Looking into the clang-win issue, one potential problem might be that at https://github.com/boostorg/thread/blob/96cd717b331e62095aa58d6768dab3baa03fcdce/src/win32/tss_pe.cpp#L219-L227 Clang issues a warning that `dw` is unused, so it's probably optimized out and the reference to `_tls_used` is eliminated.

Kojoley:
> Maybe @Kojoley knows why.

I have no idea. It crashes with access violation on trying to rethrow/read the exception from the future, reproducible only on 32bit build with optimizations.

> Clang issues a warning that dw is unused, so it's probably optimized out and the reference to _tls_used is eliminated.

When I was debugging the #316 issue I saw the symbol, it is not optimized away, and since it is marked as C extern I am not sure the comment makes any sense.

Lastique:
If it is indeed optimized out, I would call it a compiler bug. The warning does not indicate that it is removed, though.

pdimov:
This looks like a clang codegen bug. In
```
	boost::promise<int> p;
	p.set_exception( boost::copy_exception( X() ) );
```
`boost::copy_exception` returns a valid `exception_ptr`, and `set_exception` receives an invalid one. Oh well.

Kojoley:
> This looks like a clang codegen bug

I was guessing it to be also, simply because Clang has known issues with exceptions on MS ABI (https://groups.google.com/forum/#!topic/llvm-dev/ZcNUP_1550M).

Kojoley:
Weird thing, this crashes:
```cpp
#include <boost/thread/future.hpp>
#include <thread>

struct wrap
{
};

template <typename T>
boost::exception_ptr make_exception_ptr(T) {
  return boost::copy_exception(wrap{});
}

void func2(boost::promise<int> p)
{
    p.set_exception(make_exception_ptr(3));
}


int main()
{
    boost::promise<int> p;
    auto f = p.get_future();
    std::thread(func2, std::move(p)).detach();
    try
    {
        f.get();
    }
    catch (wrap const&)
    {
        return 0;
    }
    return 1;
}
```
And this does not:
```cpp
#include <boost/thread/future.hpp>
#include <thread>

struct wrap
{
};

boost::exception_ptr make_exception_ptr(int) {
  return boost::copy_exception(wrap{});
}

void func2(boost::promise<int> p)
{
    p.set_exception(make_exception_ptr(3));
}


int main()
{
    boost::promise<int> p;
    auto f = p.get_future();
    std::thread(func2, std::move(p)).detach();
    try
    {
        f.get();
    }
    catch (wrap const&)
    {
        return 0;
    }
    return 1;
}
```

```diff
-template <typename T>
-boost::exception_ptr make_exception_ptr(T) {
+boost::exception_ptr make_exception_ptr(int) {
   return boost::copy_exception(wrap{});
 }
```

pdimov:
Probably affects what's getting inlined where. This is what crashes for me:
```
#include <boost/thread/future.hpp>

struct X
{
};

int main()
{
    boost::promise<int> p;
    boost::unique_future<int> f = p.get_future();

    p.set_exception( boost::copy_exception( X() ) );

    try
    {
        f.get();
    }
    catch( X const& )
    {
    }
}
```
No need to use a thread at all. The `p.set_exception` line is miscompiled. I suppose one doesn't even need `future` here.

pdimov:
Yeah, this crashes too:
```
#include <boost/exception_ptr.hpp>

struct X
{
};

void set_exception( boost::exception_ptr p )
{
    try
    {
        boost::rethrow_exception( p );
    }
    catch( X const& x )
    {
    }
}

int main()
{
    set_exception( boost::copy_exception( X{} ) );
}
```

Lastique:
If it's caused by inlining, does `BOOST_NOINLINE` applied to `set_exception` help?

pdimov:
In that last example, changing `set_exception` to take by `const&` eliminates the crash.

But I'm not sure how much we need to care about this issue, as it only affects clang-cl 32 bit and nobody has reported it as a problem.

Lastique:
I'd say, we shouldn't wait until it hits someone, especially if we have a workaround.

pdimov:
The right thing to do is to file a bug against Clang, then maybe work around it.

Kojoley:
https://bugs.llvm.org/show_bug.cgi?id=46584

PR #365, opened 2022-03-06
Fix search path for relative includes

Flamefire:
On recent Cygwins and other systems the source files path may not be added to the include search path leading to:
> libs\thread\src\pthread\once.cpp:8:10: fatal error: ./once_atomic.cpp: No such file or directory
>    8 | #include "./once_atomic.cpp"

Add the path explicitly and add CI jobs for this (order/layout based on Boost.CI)

Required also for downstream libraries, see e.g. https://github.com/boostorg/atomic/pull/57

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,24 @@ environment:
       VARIANT: release
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
 
-    - ADDPATH: C:\cygwin\bin;
+    - FLAVOR: cygwin (32-bit)
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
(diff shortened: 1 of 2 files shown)


Flamefire:
@pdimov Seemingly this is not as trivial as thought as the tests heavily rely on relative includes. Sometimes in the form of `#include "../../../remove_error_code_unused_warning.hpp"` which can't be easily fixed with an `<include>.`.

So my proposal: Copy all headers used by tests (in the `test` folder or one of its subfolders) to `test/include` and add that folder to the include search path. Then adapt the include directives to `#include "remove_error_code_unused_warning.hpp"`

Any other idea?

Flamefire:
Turned out this was a B2 bug with slashes confusing the compiler -> Close

PR #369, opened 2022-05-04
Fix issue #366 (adding testcase)

sehe:
With reference to https://github.com/boostorg/thread/issues/366 and
https://github.com/boostorg/thread/pull/367

This fix is narrower than the linked PR in that it only prevents interruption exceptions from being raised inside the thread_guard destructor.

### DIFF
--- a/include/boost/thread/thread_guard.hpp
+++ b/include/boost/thread/thread_guard.hpp
@@ -34,6 +34,11 @@ namespace boost
     }
     ~thread_guard()
     {
+      // In "nesting" scenarios on_destructor may trigger interruption. Since
+      // we're in a destructor we can't afford that.
+      // See https://github.com/boostorg/thread/issues/366.
+      this_thread::disable_interruption di;
+
       CallableThread on_destructor;
 
       on_destructor(t_);
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -378,7 +378,10 @@ rule generate_self_contained_header_tests
           [ thread-run test_5502.cpp ]
     ;
 
-
+    test-suite issues
+    :
+          [ thread-run test_366.cpp ]
+    ;
 
     #explicit ts_conditions ;
     test-suite ts_conditions
--- a/test/test_366.cpp
+++ b/test/test_366.cpp
@@ -1,0 +1,42 @@
+#include "boost/thread/thread_guard.hpp"
+#include "boost/thread.hpp"
+#include <iostream>
+#include <boost/atomic.hpp>
+
+static boost::atomic_int g_sensor{0};
+struct sensing_joiner : boost::interrupt_and_join_if_joinable {
+  ~sensing_joiner() { ++g_sensor; }
+};
+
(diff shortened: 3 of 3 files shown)


sehe:
Need to also have a look at scope_guard and scoped_thread (thanks for the hint @elsamuko)

pdimov:
As far as I can see the minimal test just needs to be
```
#include "boost/thread/thread_guard.hpp"
#include "boost/thread.hpp"
#include <iostream>

static void inner_thread_func()
{
    boost::this_thread::sleep_for(
        boost::chrono::hours(1) );
}

static void outer_thread_func()
{
    boost::thread inner( inner_thread_func );
    boost::thread_guard<boost::interrupt_and_join_if_joinable> guard( inner );
}

static void double_interrupt()
{
    boost::thread outer( outer_thread_func );
    outer.interrupt();
    outer.join();
}

int main()
{
    std::cout << "Start" << std::endl;
    double_interrupt();
    std::cout << "End" << std::endl;
}
```
and, correspondingly,
```
#include "boost/thread/scoped_thread.hpp"
#include "boost/thread.hpp"
#include <iostream>

static void inner_thread_func()
{
    boost::this_thread::sleep_for(
        boost::chrono::hours(1) );
}

static void outer_thread_func()
{
    boost::scoped_thread<boost::interrupt_and_join_if_joinable> inner( inner_thread_func );
}

static void double_interrupt()
{
    boost::scoped_thread<boost::interrupt_and_join_if_joinable> outer( outer_thread_func );
}

int main()
{
    std::cout << "Start" << std::endl;
    double_interrupt();
    std::cout << "End" << std::endl;
}
```
I'm not sure we need to do anything about `scoped_guard`, although I suppose it won't hurt to disable interrupts in its destructor, too. There's no scenario in which you want them not disabled.

pdimov:
Now that I look at it, maybe we can replace
```
static void double_interrupt()
{
    boost::thread outer( outer_thread_func );
    outer.interrupt();
    outer.join();
}
```
with
```
static void double_interrupt()
{
    boost::thread outer( outer_thread_func );
    boost::thread_guard<boost::interrupt_and_join_if_joinable> guard( outer );
}
```
for symmetry.

elsamuko:
@sehe @pdimov `scope_guard` was just a helper class in my example code and is not part of `boost::thread`. The thread functors are currently used in
* `boost::strict_scoped_thread`
* `boost::scoped_thread`
* `boost::thread_guard`

pdimov:
Well, we can't do anything about user-defined scope guards.

pdimov:
All right, we need a third test for `strict_scoped_thread`, then.

sehe:
I'll add the test and fix for `*scoped_thread`

pdimov:
Should be fixed with https://github.com/boostorg/thread/commit/f71e0f1645413c3291e37d766e5b04bc57c6e132 and https://github.com/boostorg/thread/commit/8db325363b8e91de23c062ec8964bb605ad89f11.

PR #392, opened 2023-09-02
Switch to `boost::core::invoke_swap`

Lastique:
`boost::swap` is deprecated and will be removed. Use `boost::core::invoke_swap` as a replacement.


pdimov:
The changes to swap_T_pass.cpp don't compile, so I merged https://github.com/boostorg/thread/pull/393 instead.

PR #404, opened 2024-08-09
fix run_it rvalue assignment operator - https://github.com/boostorg/thread/issues/402

### LINKED ISSUE #402: Compilation error in future.hpp using LLVM upstream clang
After this [patch](https://github.com/llvm/llvm-project/commit/8009bbec59d1c5d47ae06c431647ebee6d886ff2) in upstream LLVM, this line in `future.hpp`
https://github.com/boostorg/thread/blob/aec18d337f41d8e3081ee65f5cf3b5090179ab0e/include/boost/thread/future.hpp#L4671
fails to compile with the following error

```
boost/boost/thread/future.hpp:4672:19: error: no member named 'that' in 'run_it<FutureExecutorContinuationSharedState>'; did you mean 'that_'?
 4672 |           that_=x.that;
      |                   ^
boost/boost/thread/future.hpp:4650:55: note: 'that_' declared here
 4650 |     shared_ptr<FutureExecutorContinuationSharedState> that_;
      |                                                       ^
``` 

From looking at the error and the code surrounding the error, is this a bug in `future.hpp` where the code should be using `x.that_` given other nearby uses of `x.that_`?

zlojvavan:
fixed by replacing "x.that" with "x.that_"

fixes #402

### DIFF
--- a/include/boost/thread/future.hpp
+++ b/include/boost/thread/future.hpp
@@ -4668,7 +4668,7 @@ namespace detail
       }
       run_it& operator=(BOOST_THREAD_RV_REF(run_it) x) BOOST_NOEXCEPT {
         if (this != &x) {
-          that_=x.that;
+          that_=x.that_;
           x.that_.reset();
         }
         return *this;


Lastique:
This is already fixed by https://github.com/boostorg/thread/pull/408.

PR #419, opened 2026-06-17
Use no-exceptions-friendly macros for try/catch block

mkruskal-google:
This fixes build breakages when -fno-exceptions is used


mkruskal-google:
@Lastique - I'm pretty sure is a 2 line typo fix, it would be nice to get `-fno-exceptions` support back in boost

Lastique:
`#include <boost/core/no_exceptions_support.hpp>` is missing for the macros.

Lastique:
Merged with added includes. Thanks.

